PR #52, opened 2019-11-08
Remove intellij files and ruby version from git

egglestn:
Add intellij files and ruby version to git ignore file

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,9 @@ config.local
 # https://github.com/bkeepers/dotenv#what-other-env-files-can-i-use
 /.env.*.local
 /.env.local
+
+# Intellij files
+.idea/
+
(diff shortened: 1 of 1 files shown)


baob:
Make sense to me :)

.gitignore
```diff
@@ -21,3 +21,9 @@ config.local
 # https://github.com/bkeepers/dotenv#what-other-env-files-can-i-use
 /.env.*.local
 /.env.local
+
+# Intellij files
+.idea/
+
+# Ruby version
+.ruby-version
```
aspiers: This is fine - we don't need .ruby-version since bundler and rvm can both pluck it from `Gemfile` instead, and one source of truth for the Ruby version is better than two :-)

PR #54, opened 2019-11-08
Rubocop

egglestn:
Integrate rubocop gem
Reconfigure rubocop to reduce the monster number of violations
Whole bunch of autofixes from rubocop, should be fairly safe but touch lots of files
Separate commits with manual updates - these look very high risk

- No problem if we decide not to merge based on risk, or might be a little safer after some spec tests are added

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 class ApplicationController < ActionController::Base
   include ApplicationHelper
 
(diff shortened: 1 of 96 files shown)


app/controllers/application_controller.rb
```diff
@@ -8,32 +10,32 @@ class ApplicationController < ActionController::Base
...
+    else
+      redirect_to root_path
     end
-    @user = current_user
+  end
+
+  def login_with
+    session[:return_to] = url_except_param(request.original_url, :log_in_with)
+    logger.debug "After login will return to #{session[:return_to]}"
+    redirect_to root_path(log_in_with: params[:log_in_with])
```
baob: I presume you did this method extraction and not rubocop ?

If I've understood it correctly I suspect it needs making private so that it does not become usable as an action in it's own right.

egglestn: :+1: good catch, thanks

egglestn: I've fixed it, but there might be others like this, will double check myself if we go ahead with PR

egglestn:
Suggest close in favour of https://github.com/swapmyvote/swapmyvote/pull/55

aspiers:
Makes sense - merged #55 so closing this.  Style fixes can be submitted as separate PRs.

PR #55, opened 2019-11-08
Add rubocop gem and config

egglestn:
Rubocop gem and config only, with no fixes - the fixes seemed pretty high risk, so this just enables developers to run and fix for themselves.

At time of commit, currently rubocop was finding 338 offenses

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,0 +1,25 @@
+AllCops:
+  Include:
+  - app/**/*.rb
+  - Rakefile
+  - '**/Gemfile'
+  - config.ru
+  Exclude:
+  - config/**/*
+  - db/**/*
+  - spec/**/*
+  - bin/**/*
+  - lib/tasks/*
+  TargetRubyVersion: 2.3
+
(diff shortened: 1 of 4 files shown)


.rubocop.yml
```diff
@@ -0,0 +1,25 @@
...
+  - Rakefile
+  - '**/Gemfile'
+  - config.ru
+  Exclude:
+  - config/**/*
+  - db/**/*
+  - spec/**/*
+  - bin/**/*
+  - lib/tasks/*
+  TargetRubyVersion: 2.3
```
baob: Maybe update this ... it may reduce the number of autocorrects.

egglestn: Fixed, but sadly no, still 338 offenses. We could increase the maxes a bit more, and catch a few more of the method length issues, for example

baob:
I'm in favour ... defaults are not all what I'd choose, but I think consistency is more important than those choices, so all good from me.

aspiers:
Great, thanks!

Issue #57, opened 2019-11-08
fix Rubocop errors


aspiers:
Since #55 was merged we now have Rubocop infrastructure for linting style inconsistencies and other syntax issues.  This can be done via:

    bundle exec rubocop

Currently there are a number of different issues.  This can potentially be tackled via divide and conquer - my main suggestion would be to stick to PR per type of style issue (i.e. per category of Rubocop error or warning).  That will make the PRs easier to review, and should help minimise the need to block merging a PR, since it will avoid mixing more controversial changes with less controversial ones.

If anyone is planning to help with aspects of this, please coordinate on Slack to avoid stepping on other people's toes.  Thanks!

baob:
I will pitch in by rubocopping any files I touch from now on.  I may add guard if that helps me with that goal.

gabrielrumiranda:
Can I take this issue ?

aspiers:
@gabrielrumiranda Your help is extremely welcome although please see my suggestion on the divide and conquer approach above.  I definitely **don't** want to have to review one giant PR which fixes every single rubocop issue at once, as that would be a huge amount of effort to get right.

gabrielrumiranda:
Only left 4 errors, #59

PR #59, opened 2019-11-08
Solving rubocop problems

gabrielrumiranda:


app/controllers/application_controller.rb
```diff
@@ -8,32 +10,33 @@ class ApplicationController < ActionController::Base
...
-      return
     end
     @user = current_user
   end
 
+  def login
+    session[:return_to] = url_except_param(request.original_url, :log_in_with)
+    logger.debug "After login will return to #{session[:return_to]}"
+    redirect_to root_path(log_in_with: params[:log_in_with])
+  end
```
baob: Is this new method an action in it's own right ? If not, I think it should be a private method.

(Did rubocop do this refactoring ? It's the 2nd time I've seen it today on a 'rubocop fixes' branch  :)

baob:
Can we really be confident that rubocop didn't break anything here, given that we have no tests. (And given that it's not unknown for rubocop to break things even with tests in place).

Maybe I'm being too cautious :-/

gabrielrumiranda:
Some functions I refactory because is so big and make more than one action, buSome functions I refactory because is so big and make more than one action, but now I go make a PR where only solve problems automatic by rubocop

gabrielrumiranda:
#60

aspiers:
I agree with @baob that some real caution is needed here to due to lack of tests.  I have explicitly detailed the way I want to address this in #57 so please read that carefully.

aspiers:
Handling this a different way.

PR #60, opened 2019-11-08
Solving automatic rubocop problems

gabrielrumiranda:


aspiers:
Handling this a different way.

PR #62, opened 2019-11-08
Add guard

baob:
Leave guard running in a separate terminal, 

    $ bundle exec guard

and it will run specs every time you save a file.

Will also run bundler when needed, and rubocop, though in a fail-fast mode, as we have so many rubcop errors right now, so you'll only ever see errors for the most recently touched bad file. 

If rspec gives errors, you won't be bothered  with rubocop notifications.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -78,6 +78,14 @@ group :development, :test do
   # Have not fixed all the issues, because many of the fixes are high risk: suggest
   # we tidy up each file as and when working on it
   gem "rubocop"
+
+  # Guard checks everything when you save a file
+  gem "guard"
+  gem "guard-rspec"
+  gem "guard-rails"
+  gem "guard-bundler"
+  gem "guard-rubocop"
+  # gem "terminal-notifier-guard"
 end
 
 group :production do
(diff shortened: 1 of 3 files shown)


baob:
Yes ... not entirely sure about guard-rails myself, never used it, but from the name it looked like a must-have. Not sure why it's spinning up webrick (or any server) particularly as we have no tests that exploit it. May be best to comment it out in the Guardfile.

aspiers:
[**@baob**](https://github.com/baob) commented on [November 9, 2019 10:47 AM](https://github.com/swapmyvote/swapmyvote/pull/62#issuecomment-552088363 "2019-11-09T10:47:55Z - Replied by Github Reply Comments"):
> Yes ... not entirely sure about guard-rails myself, never used it, but from the name it looked like a must-have. Not sure why it's spinning up webrick (or any server) particularly as we have no tests that exploit it. May be best to comment it out in the Guardfile.

Already commented out :-)  I think the whole point of that plugin is to automatically spin up the server and restart when appropriate.  Nice idea but I prefer having my server output in a different terminal to my guard test output ...

PR #63, opened 2019-11-08
Improvements to Sass/CSS, use normalize.css, basic fixes for small viewports

georgecrawford:
Smallscreen header:

Smallscreen footer:

Largescreen footer:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -39,6 +39,7 @@ gem 'omniauth-rails_csrf_protection', '~> 0.1'
 gem 'haml-rails'
 gem 'addressable'
 gem 'normalize-rails'
+gem "autoprefixer-rails"
 
 gem 'seedbank'
 
--- a/app/assets/stylesheets/_variables.scss
+++ b/app/assets/stylesheets/_variables.scss
@@ -1,7 +1,9 @@
 $base-font-size: 18px;
 $base-line-height: 1.6;
-$computed-line-height: $base-line-height * $base-font-size;
+
+$smallscreen-width: 650px;
 
 $border-color: #ddd;
-
 $button-color: #457BDA;
+$facebook-color: #3B5998;
+$twitter-color: #55ACEE;
--- a/app/assets/stylesheets/home.scss
+++ b/app/assets/stylesheets/home.scss
@@ -3,6 +3,8 @@
 @import "_variables";
 @import "_mixins";
 
+@import 'normalize-rails/normalize';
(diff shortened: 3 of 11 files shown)


app/assets/stylesheets/_variables.scss
```diff
@@ -1,7 +1,9 @@
 $base-font-size: 18px;
 $base-line-height: 1.6;
-$computed-line-height: $base-line-height * $base-font-size;
```
aspiers: We don't need this for anything?

app/assets/stylesheets/home.scss
```diff
@@ -66,7 +66,6 @@ p.choose_party {
   background-color: white;
 
   .user {
-    display: inline-block;
```
aspiers: I guess these changes are deliberate but not sure how they relate to rationalising variables - were they supposed to be in a different commit?

Gemfile
```diff
@@ -39,6 +39,7 @@ gem 'omniauth-rails_csrf_protection', '~> 0.1'
 gem 'haml-rails'
 gem 'addressable'
 gem 'normalize-rails'
+gem "autoprefixer-rails"
```
aspiers: According to https://github.com/ai/autoprefixer-rails#ruby-on-rails we need to run `rake tmp:clear` after introducing this - is that right?  If so would be helpful to mention this kind of thing in the commit message and on Slack to avoid people getting caught out when they `git pull`.

aspiers:
Addresses #5, at least partially.

aspiers:
I wrote:
> So I'm going to assign that to you for now and mark it in progress, to avoid anyone else taking it on without realising you're working on it.

Oh crap, I can't until you're added to the dev team :-(

aspiers:
Tested locally and confirmed this is a gigantic improvement for mobile devices :-) :+1:

app/assets/stylesheets/_variables.scss
georgecrawford: Nope, it wasn't used.

app/assets/stylesheets/home.scss
georgecrawford: Sorry, my commits weren't brilliantly distinct. I will try and describe things better in future.

Gemfile
georgecrawford: Thanks. I missed that. I'll ping Slack.

Issue #41, opened 2019-11-03
Maybe eliminate duplication of modal login dialogs


aspiers:
There are two identical-looking modal login dialogs:

- `.js-submit-modal` in `home/index`
- `.js-login-modal` in `layouts/application`

Both are susceptible to CVE-2015-9284 as mentioned in #19.  @jpallen Is there a good reason for needing this duplication?  Can't we scrap the one in `home/index` and just use `.js-login-modal` everywhere?

aspiers:
OK so the unification at #64 exposed why two separate ones were added.  Swap preferences need to be submitted along with the login (which is not happening with the application-wide modal, thereby explaining #36), and for mitigating CVE-2015-9284 login needs to happen as a `POST` (and therefore cannot be done via an HTTP redirect as this is impossible by design in HTTP).  So that means we need both swap preferences and the facebook/twitter login buttons within a single form.  This will require some redesign.

aspiers:
I'm trying a solution where the `<form>` is moved to the login modal, and `form="login-form"` attributes are added to the dropdown `<select>` options so that they can provide values even though they live outside the form.

Issue #36, opened 2019-11-02
Ask for preferences if logging in directly for first time


aspiers:
If a user clicks `Already been here? Log in` despite not having logged in before, their user profile will be created with the preferred / willing parties left blank, and without them being immediately asked their preferences.

aspiers:
I'm beginning to think that this is a wider problem.

aspiers:
It is - the pre-login preferences are totally ignored in all situations.

aspiers:
I confirmed 90440b2bd60d that broke saving of the user's swap preferences on login, since an `<a method="post" ...>` element inside a `<form>` does not include the other parameters within the `<form>` when submitting the action via POST.

So I think we need to replace it with a `<button>` element which does.

aspiers:
I've fixed the wider problem so this reverts back to being a corner case which is not too urgent.

baob:
Is this still a problem @aspiers ?

aspiers:
[**@baob**](https://github.com/baob) commented on [June 9, 2022 3:51 PM](https://github.com/swapmyvote/swapmyvote/issues/36#issuecomment-1151222553 "2022-06-09T14:51:53Z - Replied by Github Reply Comments"):
> Is this still a problem [@aspiers](https://github.com/aspiers) ?

I had forgotten about this but it could exactly explain the issue I saw earlier - so yes, I think it is.

baob:
@aspiers since @tdg mentioned this as urgent, I'm self-assigning so it's on my radar.

baob:
> If a user clicks `Already been here? Log in` despite not having logged in before, their user profile will be created with the preferred / willing parties left blank, and without them being immediately asked their preferences. Instead it will only ask them for their constituency via `/user/constituency/edit`, but it should actually redirect to `/user/edit`.

I can't reproduce this. It looks like, a new user created this way DOES in fact go from setting constituency to setting their voting preferences. Maybe there was a subsequent ticket that fixed this. ?

@aspiers are you able to reproduce the problem ?

baob:
> I can't reproduce this. It looks like, a new user created this way DOES in fact go from setting constituency to setting their voting preferences. Maybe there was a subsequent ticket that fixed this. ?
> 
> @aspiers are you able to reproduce the problem ?

Looks fairly important, if reproducible.

aspiers:
I will check when I get a chance.

baob:
[...]

PR #64, opened 2019-11-08
Unify social media login dialogs

georgecrawford:
The first (and the easy) part of #41.


aspiers:
Ah, now I see why this breaks things :-(  It places the modal outside the `<form>` element, and we need the form for submitting swap preferences at login time.  This also explains why #36 exists.

PR #75, opened 2019-11-11
Cookie consent

egglestn:
Display cookie consent for new users, use osana cookie consent with cdn

Cookie consent adds its own (cookie_consent status) cookie on accept, so it will need to be listed in our GDPR pages

Something (probably turbolinks) stops the cookie consent accept from firing properly until the next page refresh, so I've added some javascript with a temporary hide

API link for reference: https://www.osano.com/cookieconsent/documentation/javascript-api/
I use cookies inspector for chrome to make testing easier: https://chrome.google.com/webstore/detail/cookie-inspector/jgbbilmfbammlbbhmmgaagdkbkepnijn?hl=en

### DIFF
--- a/app/assets/javascripts/cookie_consent.coffee
+++ b/app/assets/javascripts/cookie_consent.coffee
@@ -1,6 +1,6 @@
-// The cookie consent cookie is set immediately on click, but does not take effect
-// until the next page refresh (possibly helped by turbolinks)
-// As a workaround, hide it until the user refreshes or changes page,
+# The cookie consent cookie is set immediately on click, but does not take effect
+# until the next page refresh (possibly helped by turbolinks)
+# Hide it until the user refreshes or changes page, when cookies consent should take care of it
 $ () ->
   $(document).on "click", ".cc-dismiss", ->
     console.log("Clicked dismiss")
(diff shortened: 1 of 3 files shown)


egglestn:
You shouldn't have missed anything, it should just work. One possibility is an old cookie against localhost, so check your cookies. Another is that there's a javascript bug, if so it should show up in the developer console. For managing cookies, I like https://chrome.google.com/webstore/detail/cookie-inspector/jgbbilmfbammlbbhmmgaagdkbkepnijn?hl=en

aspiers:
Yeah I saw your tip on that extension and had already installed it - great tip thanks! But I was using it and couldn't see anything obviously wrong. Presumably the cookie introduced by this PR has a unique name so there is no way an old cookie from elsewhere could interfere?  And I didn't see any errors on the console.  I even checked via the JS debugger that the cookie init code is being run.

egglestn:
The cookie name might not be unique, I will check to see. The domain will be unique in production, so the only practical chance of overlap is in dev where it is localhost. Did you restart your rails server? And if that doesn't fix it, I will need to do some head scratching. If you are on a different domain, eg lvh.me, the cookie would not save, but you should still see the prompt

egglestn:
I've pushed a commit with a better cookie name. Still not sure why it's not working for @aspiers , can you let me know what browser and OS you are using? Could it be a coffeescript issue? What did you see when you were checking that the JS fires?

aspiers:
Doh, I forgot I had the "I don't care about cookies" Chrome extension installed.

PR #85, opened 2019-11-17
Copy amends 20191117.1

tdg:
Missing and incorrect “s etc


aspiers:
This is now deployed to staging and production.

PR #89, opened 2019-11-18
Social profiles

egglestn:
Create new table for users_social_profiles with uid, provider, and nickname
Populate new table when a user signs up
Migrate existing social profile data

Summary of some related discussions

* The nickname isn't coming through in the omniauth data, need to work out why not -> due to FB blocking our requests for the profile url, will keep the field for now on assumption that we will get FB approval in future
* There is potential to change users -> identities from has_one to has_many, but this makes it hard to work out which profile_url to use, we will need a priority decision if a user has two profiles: will leave as singular for now, and delete uid, profile_url, and image_url (but not email!)
* There is no profile_url for FB, just return generic FB url for now, until we have FB approval for the user_link


db/migrate/20191118151240_migrate_profile_data.rb
```diff
@@ -0,0 +1,20 @@
+class MigrateProfileData < ActiveRecord::Migration[5.2]
+  def up
+    User.all.each do |user|
+      next if user.users_social_profile.present?
+      next if user.provider.blank?
+
+      puts "Migrating user #{user.name} for #{user.provider}"
```
egglestn: I like a hint about what stuff changed on migration, but happy to remove if it's not your style

aspiers: I like this too; keep it ;-)

db/schema.rb
```diff
@@ -73,4 +73,14 @@
     t.boolean "sent_vote_reminder_email", default: false
   end
 
+  create_table "users_social_profiles", force: :cascade do |t|
+    t.integer "user_id"
```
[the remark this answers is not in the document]
egglestn: Good catch, thanks

db/schema.rb
```diff
@@ -73,4 +73,14 @@
     t.boolean "sent_vote_reminder_email", default: false
   end
 
+  create_table "users_social_profiles", force: :cascade do |t|
```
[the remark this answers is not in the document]
egglestn: I like a shorter name, I suggest identities or profiles?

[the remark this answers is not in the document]
aspiers: Yeah `identities` sounds good to me.

app/models/users_social_profile.rb
```diff
@@ -0,0 +1,28 @@
+class UsersSocialProfile < ApplicationRecord
+  belongs_to :user
+
+  delegate :uid, to: :user, prefix: false
```
aspiers: Eventually we'll allow multiple profiles per user and will need a `uid` per profile, so can we move `uid` to this new table now?  uids are more related to a given identity provider than with the user, so this feels more logical, and I see you already included `uid` in the migration creating this table so I'm guessing you agree ;-)

app/models/users_social_profile.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  enum provider: %i[
+    twitter
+    facebook
+  ]
+
+  def self.from_omniauth(auth, user)
+    where(provider: auth.provider, uid: auth.uid, user_id: user.id).first_or_initialize.tap do |usp|
+      usp.nickname = auth.info.nickname
+      usp.save!
```
aspiers: As per above we'd store the uid in the new table here.

db/migrate/20191118142930_create_social_profiles.rb
```diff
@@ -0,0 +1,12 @@
+class CreateSocialProfiles < ActiveRecord::Migration[5.2]
+  def change
+    create_table :users_social_profiles do |t|
+      t.references :user, foreign_key: true
+      t.integer :provider
+      t.string :nickname
+      t.string :uid
```
aspiers: Need `profile_url` and `image_url` fields here.  In fact let's also include `email` here too, and delegate to it from `User`.

db/schema.rb
```diff
@@ -73,4 +73,14 @@
...
+    t.integer "user_id"
+    t.integer "provider"
+    t.string "nickname"
+    t.string "uid"
```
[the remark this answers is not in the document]
aspiers: Oh sorry, I forgot that `profile_url` is a method for now.  Don't worry, we can add that later in the `user_link` PR later.  But I think we do need `image_url` right now, don't we?

app/models/user.rb
```diff
@@ -30,16 +33,6 @@ def self.from_omniauth(auth)
...
-    return "https://facebook.com/#{uid}" \
-      if provider == "facebook"
-
-    return "https://twitter.com/intent/user?user_id=#{uid}" \
-      if provider == "twitter"
-
-    return "#"
-  end
-
   def image_url
```
aspiers: This should delegate to the new model.

db/schema.rb
```diff
@@ -73,4 +73,14 @@
...
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["user_id"], name: "index_users_social_profiles_on_user_id"
```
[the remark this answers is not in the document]
egglestn: @aspiers the migration has the t.references that generates the index in the schema, so I think it's correct as is?

db/schema.rb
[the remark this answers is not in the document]
egglestn: Yup, image_url added. I wanted to migrate more of the fields, so this is good.

app/models/user.rb
egglestn: Think it does now

app/controllers/sessions_controller.rb
```diff
@@ -9,6 +10,31 @@ def create
...
+    return user
+  end
+
+  def user_for_identity(auth)
+    identities = Identity.where(provider: auth.provider, uid: auth.uid)
+
+    return unless identities.any?
+
+    identities.first&.user
+  end
```
egglestn: It would be simpler to just look the user up directly by email address, but it's not safe: we want users to be able to edit their email address separately from their social media email address, so it might not match

app/models/user.rb
```diff
@@ -185,4 +167,23 @@ def mobile_number=(new_number)
...
+    identity&.provider
+  end
+
+  def uid
+    identity&.uid
+  end
+
+  def email
+    identity&.email
+  end
```
egglestn: Need to get rid of this, we want to have an email address on the user and it may be different to the one for the profile

aspiers:
@cla-bot check

aspiers:
Darn, Travis isn't working here either :-( Did it break I wonder?

egglestn:
Closing and re-opening to give travis a poke

aspiers:
@egglestn I've rebased this and resolved the conflicts but I can't push to this branch.  Please can you check the ["Allow edits from maintainers" checkbox](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork)?

egglestn:
"Allow edits from maintainers" was already checked, so instead we cloned Adam's changes into my branch and I have repushed. Thanks for the rebase and git support @aspiers

db/migrate/20191125151240_migrate_profile_data.rb
```diff
@@ -0,0 +1,23 @@
+class MigrateProfileData < ActiveRecord::Migration[5.2]
+  def up
+    User.all.each do |user|
+      next if user.identity.present?
+      next if user.provider.blank?
```
aspiers: This will cause all users to be skipped, because it's using your updated `#provider` method which delegates to `Identity`.

aspiers: I think we can just kill this line?  If `user.identity.present?` is false then we always want to migrate the user, right?

aspiers: I was wrong - that will migrate test users.  Found a better way and pushed a new commit to fix.

db/schema.rb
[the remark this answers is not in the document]
aspiers: You were right sorry!

Issue #20, opened 2019-01-02
Embeddable widget


tdg:
A widget would allow other third party (democracy and election tech) platforms and websites to embed on their site a front end to the vote swapping process. 

At present this could be the three questions and dropdowns, but if and as the SmV platform gets more sophisticated, it could remain with basic questions like these, directing user/voters to the next part of the process.

(Inspiration includes @democracyclub https://github.com/DemocracyClub/WhereDoIVote-Widget etc)

PR #92, opened 2019-11-19
don't prevent deep-linking

aspiers:
This seems a bit draconian; unless there is a really good reason
let's drop it.

### DIFF
--- a/app/views/static_pages/terms.html.haml
+++ b/app/views/static_pages/terms.html.haml
@@ -177,9 +177,7 @@
 
     %p
 
-      Our website must not be framed on any other website, nor may you
-      create a link to any part of our website other than the home
-      page.
+      Our website must not be framed on any other website.
 
     %p
 


tdg:
I'm fine with this and don't know why we went for this approach originally - @jpallen might be able to elucidate?

We did discuss in summer 2017 #20 which would create an embeddable widget presumably for use (framing?) in other websites, so even that part of it would be under review if we progress that Issue.

jpallen:
Also fine with this. The terms don't look like something I wrote, possibly it's boilerplate from somewhere?

egglestn:
Yes it was boiler plate. I grabbed sample templates from a range of sources, rephrased them slightly for consistency, and took out everything that was definitely wrong. I left in a few of the marginal paragraphs (including this one) in case they reminded us of something relevant we need to take care of. Happy for them to be tidied up and removed, suggest you look round in case there are any more.

app/views/static_pages/terms.html.haml
```diff
@@ -177,9 +177,7 @@
 
     %p
```
egglestn: Should probably take out this %p as well, not a blocker

jpallen:
That's why I don't recognise! It was only added yesterday :)

PR #100, opened 2019-11-21
Update branding and background for 2019

### LINKED ISSUE #96: Update logos & bg for 2019
Recreate files in `app/assets/images` using .psd linked below
https://www.dropbox.com/s/utoof2gqguomlm3/SwapMyVote-logos-2019.psd?dl=0

NB there are three layers - square, vertical, and inline.

We also need to use a new background image, as the source of our original is unknown

tdg:
More dynamic and eye grabbing, friendly background (for which we have
the rights)

Resolves #96


tdg:
Before

FireShot Capture 007 - Swap My Vote - swapmyvotedev herokuapp com

tdg:
After

Swap My Vote - localhost - mobresponse

aspiers:
I didn't merge the new `pattern.png` as it was too distracting, even (IMHO) when faded.  We can do that separately.

PR #104, opened 2019-11-22
Redact names

### LINKED ISSUE #47: Protect voters' data better before swap - redact part of name
Screen Shot 2019-11-06 at 18 37 04
Ideal would be to keep first name and second initial. 

However, names don't all follow the same rules, so could we redact the last `50%-1` of characters? Better ideas welcome

threedaymonk:
This:

- Redacts names as described in #47, paying attention to potential edge cases and non-ASCII names
- Exposes a `redacted_name` attribute on the `User` model (with an eye towards letting people set their on redacted name in future)
- Changes references to `name` in the views to use `redacted_name`

I think I've captured all the locations in which the name is displayed before a swap is confirmed, but it would be good if someone who is more familiar with the app could check that I (a) haven't missed any, and (b) haven't been too assiduous and redacted names where they shouldn't be.

Closes #47

### DIFF
--- a/app/lib/name_redactor.rb
+++ b/app/lib/name_redactor.rb
@@ -1,0 +1,32 @@
+require "singleton"
+require "forwardable"
+
+class NameRedactor
+  include Singleton
+  extend SingleForwardable
+
+  def_delegator :instance, :redact
+
+  def redact(name)
+    return nil unless name
+
+    first, second, _rest = split_names(strip_leading_blanks(name))
+    [first, first_grapheme(second)].compact.join(' ')
+  end
+
+  private
+
+  def split_names(str)
+    str.split(/[[:space:]]+/)
+  end
+
+  def first_grapheme(str)
+    return nil unless str
+
+    str.each_grapheme_cluster.first
+  end
+
+  def strip_leading_blanks(str)
+    str.sub(/\A[[:space:]]+/, '')
+  end
+end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -167,4 +167,8 @@ class User < ApplicationRecord
     save
     UserMailer.reminder_to_vote(self).deliver_now
   end
+
+  def redacted_name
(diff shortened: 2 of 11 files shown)


threedaymonk:
> One question though - is there any reason `name_redactor.rb` shouldn't live in `lib/` which already exists, rather than creating a new `app/lib/` directory?

Yes: everything under `app` is autoloaded, while autoloading from lib is [discouraged in new Rails apps](https://guides.rubyonrails.org/autoloading_and_reloading_constants_classic_mode.html):

> But using autoload_paths on its own in the past (before Rails 5) developers might configure autoload_paths to add in extra locations (e.g. lib which used to be an autoload path list years ago, but no longer is). However this is now discouraged for most purposes, as it is likely to lead to production-only errors. It is possible to add new locations to both config.eager_load_paths and config.autoload_paths but use at your own risk.

I would have given it a more specifically named directory, but I spent too much time thinking about what that should be so went with `lib` as a default.

aspiers:
Ah OK thanks!  I should have known that since funnily enough I was reading the exact same docs the other day when figuring out where to put [a new class for sending SMS messages via MessageBird](https://github.com/swapmyvote/swapmyvote/compare/master...aspiers:phone#diff-770efdf366ab4cf37339747a35bfedcd).

Another question (since your Rails knowledge is clearly more up to date than mine): what's the advantage of using `SingleForwardable` over `extend self`?

aspiers:
For example, this works fine:

```diff
1 file changed, 2 insertions(+), 8 deletions(-)
app/lib/name_redactor.rb | 10 ++--------

modified   app/lib/name_redactor.rb
@@ -1,11 +1,5 @@
-require "singleton"
-require "forwardable"
-
-class NameRedactor
-  include Singleton
-  extend SingleForwardable
-
-  def_delegator :instance, :redact
+module NameRedactor
+  extend self
```

PR #107, opened 2019-11-22
use extend self instead of SingleForwardable

aspiers:
This is simpler and achieves the same effect.  However if there is an advantage of `SingleForwardable` I'm missing then please let me know, as I'm very happy to learn something new and ditch this PR ;-)

### DIFF
--- a/app/lib/name_redactor.rb
+++ b/app/lib/name_redactor.rb
@@ -1,11 +1,5 @@
-require "singleton"
-require "forwardable"
-
-class NameRedactor
-  include Singleton
-  extend SingleForwardable
-
-  def_delegator :instance, :redact
+module NameRedactor
+  extend self
 
   def redact(name)
     return nil unless name


aspiers:
Hmm, well Hound CI doesn't like it.  `module_function` would be needed on all the private methods too, which is clearly worse than the current approach.  Interesting!

threedaymonk:
I think the advantages of `SingleForwardable` are:

1. It's more explicit and focused in what it does, whereas with `extend` you're relying more on a side-effect
2. It preserves the encapsulation of private methods (as you've just found out!)

aspiers:
[**@threedaymonk**](https://github.com/threedaymonk) commented on [November 22, 2019 12:49 PM](https://github.com/swapmyvote/swapmyvote/pull/107#issuecomment-557519668 "2019-11-22T12:49:44Z - Replied by Github Reply Comments"):
> I think the advantages of `SingleForwardable` are:
> 
> 1. It's more explicit and focused in what it does, whereas with `extend` you're relying more on a side-effect

Agreed.  OTOH, most Ruby devs know what `extend self` does whereas this approach seems more obscure.  Maybe that's just me, but I had to look it up so I'm guessing others would need to too.

> 2. It preserves the encapsulation of private methods (as you've just found out!)

So does `extend self` AFAICS; using the branch of this PR:

```
$ rails c
Loading development environment (Rails 5.2.3)
2.6.5 :001 > NameRedactor
 => NameRedactor 
2.6.5 :002 > NameRedactor.redact("Adam Spiers")
 => "Adam S" 
2.6.5 :003 > NameRedactor.split_names("Adam Spiers")                                                                                                           
Traceback (most recent call last):
        1: from (irb):3
NoMethodError (private method `split_names' called for NameRedactor:Module)
```

threedaymonk:
> So does extend self AFAICS; using the branch of this PR:

Ah yes, it's `module_function` that doesn't work with private methods.

I think you can choose any two of:

1. Preserve private method encapsulation
2. Get rid of `SingleForwardable`
3. Keep the Rubocop/Hound configuration that doesn't like `extend self`

Another option would be to replace `def_delegator` with a proxy method:

```ruby
def self.redact(name)
  instance.redact(name)
end
```

Or you could get rid of `Singleton` too if it seems too obscurantist:

```ruby
def self.redact(name)
  new.redact(name)
end
```

I think it's up to you!

aspiers:
Let's keep it as it is, but thanks a lot for helping me learn some new Ruby today! 😁

PR #111, opened 2019-11-22
Correct name redactions

### LINKED ISSUE #110: Check whether swap partner names are redacted in the right places.
I merged #104 which was a beautiful implementation of name redaction to address #47.  However I might have been a little too hasty in merging since @threedaymonk wrote:

> I think I've captured all the locations in which the name is displayed before a swap is confirmed, but it would be good if someone who is more familiar with the app could check that I (a) haven't missed any, and (b) haven't been too assiduous and redacted names where they shouldn't be.

and I didn't do that prior to merging.

Taking a look now, I see that names are [redacted when a swap is confirmed by email](https://github.com/swapmyvote/swapmyvote/blob/master/app/views/user_mailer/swap_confirmed.html.haml), but [not via the web UI](https://github.com/swapmyvote/swapmyvote/blob/master/app/views/user/vote/show.html.haml), nor [when a swap is cancelled by email](https://github.com/swapmyvote/swapmyvote/blob/master/app/views/user_mailer/swap_cancelled.haml).  There may be other inconsistencies, so we need a full review of this.

@tdg Based on how you worded #47, it sounds like you envisioned redaction only applying *prior* to swaps being confirmed; is that right?  If so we can [unredact](https://www.youtube.com/watch?v=2aINa6tg3fo) the confirmation email.

@threedaymonk If this isn't too boring, perhaps you could run with it?

threedaymonk:
Closes #110

### DIFF
--- a/app/views/user_mailer/swap_confirmed.html.haml
+++ b/app/views/user_mailer/swap_confirmed.html.haml
@@ -2,13 +2,13 @@
 	Hi #{@user.name},
 %p
 	You're all set to swap your vote with
-	%a{href: @swap_with.profile_url} #{@swap_with.redacted_name}!
-	#{@swap_with.redacted_name} is going to vote for the #{@swap_with.willing_party.name} party for you, in the #{@swap_with.constituency.name} constituency.
+	%a{href: @swap_with.profile_url} #{@swap_with.name}!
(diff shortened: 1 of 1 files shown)


aspiers:
@threedaymonk Do I remember right that you're back in action on Monday, and that this would be top of your list? Just doing a bit of project planning at the mo - cheers ;-)

Issue #109, opened 2019-11-22
Need better test coverage


aspiers:
Following on from #39 and #105 we're now in a better place with tests, but still I suspect our coverage is very low.

First step is to automate measuring coverage; I've filed #108 for that.  Once we have automatic measurement, we can start working on improving the coverage, which is what this issue should be used for tracking.

baob:
I could be way out of date on this, but in the past I've used SimpleCov and tied it in to Guard ... though guard is probably best applied when we have near to 100% coverage (say 80/90 ?) to keep us on track, rather than getting us to that point.

aspiers:
Continuing tools discussion in #108 ...

PR #124, opened 2019-11-23
Fix API and add more tests

aspiers:
Fix a bug which was breaking the API, and add a load of tests.

Contributes towards #109.


spec/support/user_sessions.rb
```diff
@@ -0,0 +1,11 @@
...
+  let(:user) { create(:user) }
+
+  before do
+    session[:user_id] = user.id
+  end
+end
+
+RSpec.configure do |rspec|
+  rspec.include_context "user logged in", logged_in: true
+end
```
baob: Nice :)

PR #136, opened 2019-11-23
separate SERVER_HOST from port so we can choose a cookie name correctly

### LINKED ISSUE #134: cookie consent is broken on staging and production
Something broke :-( and now the consent cookie isn't being stored on staging and production, but it is working on localhost?!  Maybe an issue with having multiple domains in the Javascript?

aspiers:
Previously we were only using SERVER_HOST to construct URLs in emails we send.  Now we want to use it to decide the right cookie name, so store the port number (if needed) in a separate variable.

Note that this requires every developer to tweak their `.env.development.localhost` file to match `.env.example`!

Fixes #134.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,7 +1,11 @@
+# Determine what cookie name to use.  This can also be used to
+# construct SERVER_HOST_PORT below.
+export SERVER_HOST=localhost
+
 # This is used to figure out what URLs to generate when sending out
 # email.  It can't be done dynamically because the mailer instance
 # doesn't have any context about the incoming request.
-export SERVER_HOST_PORT=localhost:3000
+export SERVER_HOST_PORT=$SERVER_HOST:3000
 
 # You will need a Facebook application to allow log-ins via Facebook.
 # Obtain the app key and secret from https://developers.facebook.com/apps/
--- a/app/views/layouts/_cookie_consent.html.haml
+++ b/app/views/layouts/_cookie_consent.html.haml
@@ -12,10 +12,7 @@
       },
       cookie: {
         name: "_swapmyvote_cookie_consent",
-        domain: ".swapmyvote.uk",
-        domain: ".swapmyvote.herokuapp.com",
-        domain: ".swapmyvotedev.herokuapp.com",
-        domain: "localhost"
+        domain: "#{ENV['SERVER_HOST']}"
       }
     });
   });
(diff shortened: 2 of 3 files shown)


app/views/layouts/_cookie_consent.html.haml
```diff
@@ -12,10 +12,7 @@
       },
       cookie: {
         name: "_swapmyvote_cookie_consent",
-        domain: ".swapmyvote.uk",
-        domain: ".swapmyvote.herokuapp.com",
-        domain: ".swapmyvotedev.herokuapp.com",
-        domain: "localhost"
+        domain: "#{ENV['SERVER_HOST']}"
```
aspiers: This might cause a regression without the `.` prefix - @egglestn what do you think?

egglestn: The . prefix is recommended for subdomains, and I think also helps with older browser support. If we don't have any subdomains, it probably works, I don't have a strong opinion on whether we keep it or not - would have to test to be sure.

egglestn:
Apologies, I missed this statement in the release docs "Each site configuration may now have only one domain" - that'll be why the multiple domain list doesn't work. There are examples in stack overflow showing multiple domains with the syntax you've just removed, but I now realise must be from earlier releases

.env.example
```diff
@@ -1,7 +1,11 @@
+# Determine what cookie name to use.  This can also be used to
+# construct SERVER_HOST_PORT below.
+export SERVER_HOST=localhost
+
 # This is used to figure out what URLs to generate when sending out
 # email.  It can't be done dynamically because the mailer instance
 # doesn't have any context about the incoming request.
-export SERVER_HOST_PORT=localhost:3000
+export SERVER_HOST_PORT=$SERVER_HOST:3000
```
egglestn: Is this always 3000 in heroku? I know in AWS you can't guarantee the server port and should pick that from a variable too

aspiers: No, it's port 80 in Heroku (i.e. no port needed).  This is just the example file for developers; I take care of setting the Heroku environments myself via `heroku config:set`.

app/views/layouts/_cookie_consent.html.haml
aspiers: We only use `www.swapmyvote.uk` and `swapmyvotedev.herokuapp.com` - no sub-domains, so I guess we're good here.

PR #146, opened 2019-11-25
delete unused test/ hierarchy

aspiers:
Make sure we don't confuse people into thinking we have no tests.


egglestn:
Could you remove the entire test folder?

aspiers:
@egglestn That's what this should do, since git doesn't track directories and this removes all the files under `test/`.

PR #160, opened 2019-11-26
Add client-side phone number validation

### LINKED ISSUE #132: Client-side validation of phone number
Probably best candidate: [jackocnr/intl-tel-input: A JavaScript plugin for entering and validating international telephone numbers](https://github.com/jackocnr/intl-tel-input)

These also look useful:

- [mobi/telephone_number: Phone number validation for Ruby](https://github.com/mobi/telephone_number)
- [rails/webpacker: Use Webpack to manage app-like JavaScript modules in Rails](https://github.com/rails/webpacker)
- [fnando/normalize_attributes: Sometimes you want to normalize data before saving it to the database like down casing e-mails, removing spaces and so on. This Rails plugin allows you to do so in a simple way.](https://github.com/fnando/normalize_attributes)

stevebaxter:
There are two main changes here:

1. Webpacker has been added to the project. This allows us to use npm modules and polyfills like Babel, with webpack bundling all the resources.
2. We add [intl-tel-input](https://www.npmjs.com/package/intl-tel-input) to do client-side phone number validation.

A few notes/caveats:

1. This introduces a new dependency on Yarn (Webpacker uses this to manage npm packages). I don't know whether we will need to do anything in deployment to make this work.
2. Most of the changes were from the webpacker install script, I'll highlight the changes that aren't part of the defaults.
3. At the moment we're bundling intl-tel-input - it's quite a large library (~400kb) but should compress to ~69kb in production. The large bit is the validation which relies on a Google library. This will be served from our domain, we could possibly switch to a CDN for this if it's going to be a problem.

Closes #132.


README.md
```diff
@@ -37,7 +37,7 @@ You will need Ruby installed.
     If this fails with error messages that mention `pg` or PostgreSQL, then
     try instead:
 
-        bundle install --without-production
+        bundle install --without=production
```
stevebaxter: I found this was wrong when I was following the instructions like a complete Ruby noobie!

app/controllers/users_controller.rb
```diff
@@ -50,6 +50,6 @@ def user_params
   end
 
   def phone_param
-    (params[:mobile_phone] || {})[:number]
+    (params[:mobile_phone] || {})[:full]
```
stevebaxter: The validation lib is set up to add a second hidden input that will contain the internationalised number, so 07856 253 263 will be turned (for the UK) into +447856253263 and stashed in this field on submit. This change allows us to capture the value from that field in the form, so that's what we store in the db.

The validation lib will reverse this and format the number nicely when we display it again.

app/javascript/packs/intlTelInput.js
```diff
@@ -0,0 +1,40 @@
+import intlTelInput from "intl-tel-input";
+
+import "intl-tel-input/build/css/intlTelInput.css";
+import "intl-tel-input/build/js/utils.js";
+
+// Set up all phone number inputs with intl-tel-input. We will add a hidden field
+// that allows forms to submit the full number. We use HTML5 validation.
+$(document).ready(() => {
```
stevebaxter: This is our code. We can write ES6 and webpack/Babel will transpile it to what's supported by our target browsers. We should be able to use TypeScript as well (my preferred option), but one step at a time.

app/javascript/packs/intlTelInput.js
```diff
@@ -0,0 +1,40 @@
...
+  const phoneInputs = document.querySelectorAll("input[type=tel]");
+  Array.from(phoneInputs).forEach(phoneInput => {
+    const intlTelPlugin = intlTelInput(phoneInput, {
+      initialCountry: "gb",
```
stevebaxter: Default to UK with UK at the top of the list.

app/javascript/packs/intlTelInput.js
```diff
@@ -0,0 +1,40 @@
...
+        const numberType = intlTelPlugin.getNumberType();
+        if (
+          numberType === intlTelInputUtils.numberType.MOBILE ||
+          numberType === intlTelInputUtils.numberType.FIXED_LINE_OR_MOBILE
+        ) {
+          // Looks valid!
+          phoneInput.setCustomValidity("");
+        } else {
+          // Valid but not a mobile number
+          phoneInput.setCustomValidity(
```
stevebaxter: We use HTML5 custom form validation, this works on IE11+. I've tested it on IE11, Chrome, Mobile Chrome, Edge, Safari, Mobile Safari.

app/views/users/edit.html.haml
```diff
@@ -30,8 +32,7 @@
 
       %p
         My phone number is
-        = telephone_field_tag "mobile_phone[number]", @mobile_number,
-                              placeholder: "07123 456 789"
+        = telephone_field_tag "mobile_phone[number]", @mobile_number
```
stevebaxter: Let the validation lib show a placeholder that's appropriate to the selected country.

README.md
aspiers: Good catch, thanks! :laughing:

app/controllers/users_controller.rb
aspiers: Thanks a lot for the explanation - hugely appreciated!  BTW in future might be helpful to actually stick it in a comment in the code since people are unlikely to come back to this PR.  So just to confirm I understand, we'll store the normalized version in the db not the user's version?

app/javascript/packs/intlTelInput.js
aspiers: Makes sense - thanks!

app/views/users/edit.html.haml
aspiers: Makes sense.

stevebaxter:
Hmm, the Hound issues should all be resolved now I've allowed ES6, (JSHint passes locally). It still says 2 are open though. I'll just try closing and reopening this to see if I can give it a kick.

README.md
```diff
@@ -24,7 +24,7 @@ sections below if you'd like to help.
 
 ## Setting up Swap My Vote locally
 
-You will need Ruby installed.
+You will need Ruby and [Yarn](https://yarnpkg.com/lang/en/docs/install/) installed.
```
aspiers: Let's explicitly mention the requirement on node 10 if that's the case.  Seems that we also need to trigger a `yarn install` somehow e.g. `rails yarn:install`?

app/controllers/users_controller.rb
stevebaxter: Yep, the backend will be presented with the normalised version. It's only a weak validation of course, as the client could always modify the JS to do anything they want.

stevebaxter: I've added a comment, I should have done that in the first place.

PR #165, opened 2019-11-27
Delete prepopulated fields from session once they've been used

aspiers:
This means that prepopulation only happens the first time the link is followed, and if the user reloads the home page or returns to it later as normal (not via the API) then prepopulation will not occur.


aspiers:
@tdg Is this an improvement do you think?

aspiers:
Fair enough!

PR #169, opened 2019-11-27
Issue #159 welcome email

### LINKED ISSUE #159: updating user can cause welcome mail to be resent
## Problem description

In certain cases when user details are updated, a welcome email gets sent.  Currently this is independent of when the user's `email` field is set, and instead actually dependent on voting preferences and constituency:

https://github.com/swapmyvote/swapmyvote/blob/00b0b176fa6d69b6ab4752c953952a6a29ab326c/app/models/user.rb#L16

https://github.com/swapmyvote/swapmyvote/blob/00b0b176fa6d69b6ab4752c953952a6a29ab326c/app/models/user.rb#L148-L158

We discovered this welcome email code accidentally got triggered while running the `user.update!` in the migration in #143 which populates the `constituency_ons_id` column:

https://github.com/swapmyvote/swapmyvote/pull/143/files#diff-7b8b7e2dcd0b9eafdeccff6b739a1d3eR10

This is because that PR changes the `ready_to_swap?` method to check `constituency_ons_id` rather than `constituency_id`.  So any existing users which have `constituency_id` set but not `constituency_ons_id` prior to the migration would get the latter set by the migration, and that would trigger the email.

## Proposed solution

The `ready_to_swap?` method is a bit of a misnomer as can be seen by its code.  Also it makes more sense to send the welcome email as soon as we get an address, regardless of whether we know their voting preferences or constituency - the latter is not guaranteed, e.g. due to #36.  So @baob and I discussed and think that a more robust approach is:

- Ditch `ready_to_swap?` method.
- Add a new `SentEmail` model which `belongs_to :user` and has a `template` column.
- Every time a welcome email is sent, append a row to this table with the `template` field set to `welcome_email` tracking the delivery.  (Later we can track deliver other emails if we want)
- Replace

https://github.com/swapmyvote/swapmyvote/blob/00b0b176fa6d69b6ab4752c953952a6a29ab326c/app/models/user.rb#L16

with (caution: untested code!)

    after_save :send_welcome_email, if: needs_welcome_email?

    ...

    def needs_welcome_email?
      return !email.blank? && SentEmail.find_by(user: self, template: 'welcome_email')
    end

to ensure that the welcome is only sent to users once, as soon as we have an email for them.

baob:
Closes #159.

### DIFF
--- a/spec/models/sent_email_spec.rb
+++ b/spec/models/sent_email_spec.rb
@@ -1,0 +1,6 @@
+require "rails_helper"
+
+RSpec.describe SentEmail, type: :model do
+  specify { expect(subject).to respond_to(:user)}
+  specify { expect(subject).to respond_to(:template)}
+end
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -161,33 +161,59 @@ RSpec.describe User, type: :model do
     end
   end
 
-  describe "#send_welcome_email" do
-    let(:an_email) { double(:an_email)}
-    before do
-      allow(an_email).to receive(:deliver_now)
-      allow(UserMailer).to receive(:welcome_email).and_return(an_email)
-      subject.save!
-    end
-
+  describe "#send_welcome_email (after_save callback)" do
     context "when the user does NOT have an email address"  do
-      it "does not send an email" do
-        expect(UserMailer).not_to receive(:welcome_email)
-        subject.send_welcome_email
+      before do
+        subject.update(email: nil)
+        subject.save!
       end
 
-      specify { expect { subject.send_welcome_email }.not_to change(subject, :needs_welcome_email?).from(false) }
+      specify "#save does NOT call #send_welcome_email" do
+        is_expected.not_to receive(:send_welcome_email)
+        subject.sent_emails.clear
+        subject.save!
+      end
     end
 
     context "when the user DOES have an email address"  do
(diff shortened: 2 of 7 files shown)


spec/models/user_spec.rb
```diff
@@ -158,4 +155,60 @@
...
+        context "when NO previous welcome emails have been sent" do
+          before { subject.sent_emails.clear }
+
+          it "sends an email" do
+            expect(an_email).to receive(:deliver_now)
+            expect(UserMailer).to receive(:welcome_email).with(subject).and_return(an_email)
+            subject.save!
+          end
+
+          specify { expect { subject.save! }.to change(subject, :needs_welcome_email?).from(true).to(false) }
```
aspiers: Again Hound and rubocop should be moaning about long lines here?

baob: That one is not long enough to trigger the rule ... do you want to (separately, maybe) amend  the rule ?

aspiers: Nah, wait until #172 for that.

spec/models/sent_email_spec.rb
```diff
@@ -0,0 +1,6 @@
+require "rails_helper"
+
+RSpec.describe SentEmail, type: :model do
+  specify { expect(subject).to respond_to(:user)}
```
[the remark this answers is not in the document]
aspiers: Answered in #172.

Issue #154, opened 2019-11-25
Refresh electoral calculus poll data


baob:
Currently showing 2017 data ... new 2019 data available.

https://www.electoralcalculus.co.uk/orderedseats.html

aspiers:
Any subsequent refresh has to be manual, but, unless Electoral Calculus change their html radically, the next one should just be a cut’n’paste of their html into a spreadsheet, saved to db/fixtures.

Issue #9, opened 2018-02-06
Integrate data and predictions from opinion polls


tdg:
Integrate a regularly and automatically updated sources of predictions from opinion polls.

tdg:
See also #8

aspiers:
We don't want to get into the business of aggregating opinion polls or holding data on them, so this should be mostly a question of delegating or using available APIs.

tdg:
We will probably continue to use https://www.electoralcalculus.co.uk/orderedseats.html as a source. We need to remember to update it - we'll be prompted by EC's email alerts however

aspiers:
This isn't entirely easy until it's made more clear exactly what needs to be done.

aspiers:
I think @tdg, @baob and I managed to clarify this yesterday.  We're gonna switch to using https://www.livefrombrexit.com/tacticals/data/recommendations.json as soon as #8 is resolved, or perhaps at the same time as resolving it.

baob:
Time to clarify requirements, @tdg @aspiers ... how much of this do we actually need to show on SMV ?

I can see some obvious ones, 
* recommendation site (maybe 6-8 entries, best 4 britain etc)
* recommendations, keyed by constituency ons id and recommendation site. And containing links to our parties

![image](https://user-images.githubusercontent.com/55932/69492909-09313f00-0ea0-11ea-82a4-137af4054659.png)

aspiers:
Just merged #173.  @baob I think you had another branch up your sleeve which adds the recommendations to the frontend, right?  I'm ready for that now!

baob:
I did start on one, but subsequent edits to the that view or the CSS wrecked the display completely  so I reverted it.  https://github.com/baob/swapmyvote/commit/2e1506f8de16237dd2597c99881c0a6b55e4bbda

Now covered by this ticket --> https://github.com/swapmyvote/swapmyvote/issues/81 and this ticket https://github.com/swapmyvote/swapmyvote/issues/161 ... I think

aspiers:
OK well feel free to submit a draft PR - even if it totally balls up the display we have frontend guys who could work on that if you prefer not to :grin:

aspiers:
I think we got confused somewhere along the way - this issue is somewhat vague but was originally about opinion polls not tactical recommendations, which is covered separately in #161.  I just merged #178 which I think closes both this and #154.

PR #178, opened 2019-11-27
Issue #154 refresh polls data

baob:


aspiers:
How does this relate to the other PR?

baob:
> How does this relate to the other PR?
> […](#)
Not sure what you're asking.  This is PR #178 for Issue #154 and this is about a refresh of the polling data from electoral calculus

I have another PR #173 which is about Issue #9 and that's about integrating the aggregated recommendation from the livefrombrexit site which is a totally new thing. Though, actually, from the  title, they look confusable ... but see [this comment](https://github.com/swapmyvote/swapmyvote/issues/9#issuecomment-555463163)

db/fixtures/electoral_calculus_constituencies_tsv.rb
```diff
@@ -0,0 +1,86 @@
...
+require_relative "constituencies_original_with_ons_csv"
+
+class ElectoralCalculusConstituenciesTsv
+  include Enumerable
+
+  TSV_FILE_NAME = "./db/fixtures/electoral_calculus_constituencies.tsv"
+
+  def initialize
+    @parties_by_party_code = {
+      "con"    => Party.find_by(name: "Conservatives"),
```
aspiers: This is kind of duplicating hardcoded data already in `seeds.rb`; can we factor them both out into `db/seeds/parties.rb` which just defines a `PARTIES` `Hash` mapping the abbreviation to `[full_name, colour]` or something like that?  However I'm OK with doing that in a follow-up PR.

db/migrate/20191126092838_add_brexit_party.rb
```diff
@@ -0,0 +1,10 @@
+class AddBrexitParty < ActiveRecord::Migration[5.2]
+  def up
+    Party.find_or_create_by(name: "Brexit Party", color: "#5bc0de")
+  end
+
+  def down
+    bxp = Party.find_by(name: "Brexit Party")
+    bxp.destroy unless bxp.nil?
```
aspiers: This line is strangely pleasing to read.

aspiers:
Looks great but needs a rebase.

db/fixtures/electoral_calculus_constituencies_tsv.rb
baob: I spent a good while trying to figure out a neat solution to that, and that's what I came up with. Thing is, the mapping "con"   <=> "Conservatives" (which is essentially what's going on here) is a concern purely of this file and of nothing else. It has no purpose outside this file. The name "Conservatives" does ... but not the short code "con" which is just a convenience in the code that picks out the data from the columns of the csv/tsv. 

I'd say the refactoring could go 2 ways.
1) If we have any other use for the short code, put it in the database table.
2) if we decide it's ungainly, we might replace the code further down with
```
{
"Conservatives" => data[9],  ...
```
or whatever and eliminate the shortcode entirely.

aspiers: > Thing is, the mapping "con" <=> "Conservatives" (which is essentially what's going on here) is a concern purely of this file and of nothing else. It has no purpose outside this file.

Not sure I follow - the exact same mapping is needed in `seeds.rb` below.

aspiers: But appeasing my OCD tendencies for following DRY is not nearly as important as other stuff on the list, so probably just forget about it for now ;-)

baob: > the exact same mapping is needed in seeds.rb below
In that case my intent to delete it (because I know it's not used) was forgotten.

I will fix it now.

aspiers: OK thanks!

PR #179, opened 2019-11-27
Fix webpack CSS for production sites

stevebaxter:
We need to include a `stylesheet_pack_tag` in the haml for each pack we use in that view. This will cause the right CSS to get included. This isn't a problem for dev builds because there the CSS is inlined in the JS for live reloading.


aspiers:
Works on staging environment, thanks a lot!

PR #180, opened 2019-11-27
Issue #125 swap expiry config

### LINKED ISSUE #125: Make swap expiry time configurable
Need to parameterise the swap expiry time via an environment variable.  It is mentioned in the FAQ, a few email and web templates, and also `lib/tasks/swaps.rake`.

baob:
Closes #125.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -36,7 +36,7 @@ class User < ApplicationRecord
 
   def potential_swap_users(number = 5)
     # Clear out swaps every few hours to keep the list fresh for people checking back
-    potential_swaps.where(["created_at < ?", DateTime.now - swap_validity_period]).destroy_all
+    potential_swaps.where(["created_at < ?", DateTime.now - 2.hours]).destroy_all
     create_potential_swaps(number)
     swaps = potential_swaps.all.eager_load(
       target_user: { constituency: [{ polls: :party }] }
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -24,6 +24,8 @@ class User < ApplicationRecord
   after_save :send_welcome_email, if: :needs_welcome_email?
   before_destroy :clear_swap
 
+  include ApplicationHelper
+
   def omniauth_tokens(auth)
     self.token = auth.credentials.token
     if auth.credentials.expires_at
@@ -34,7 +36,7 @@ class User < ApplicationRecord
 
   def potential_swap_users(number = 5)
     # Clear out swaps every few hours to keep the list fresh for people checking back
-    potential_swaps.where(["created_at < ?", DateTime.now - 2.hours]).destroy_all
+    potential_swaps.where(["created_at < ?", DateTime.now - swap_validity_period]).destroy_all
     create_potential_swaps(number)
     swaps = potential_swaps.all.eager_load(
       target_user: { constituency: [{ polls: :party }] }
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -24,8 +24,6 @@ class User < ApplicationRecord
   after_save :send_welcome_email, if: :needs_welcome_email?
   before_destroy :clear_swap
 
-  include ApplicationHelper
-
   def omniauth_tokens(auth)
     self.token = auth.credentials.token
     if auth.credentials.expires_at
--- a/app/views/user_mailer/welcome_email.html.haml
+++ b/app/views/user_mailer/welcome_email.html.haml
@@ -19,7 +19,7 @@
 		%li If need be, check that there is a candidate standing there (there is a link below)
 		%li <em>Click your preferred voting partner to propose a vote swap!</em>
 
-	We want to find as many people as possible a good swap, so, for the final few days of the campaign, we are shortening the time for people to confirm a swap, down to just <em>*#{swap_validity_hours} hours *</em>. If the swap is not confirmed within #{swap_validity_hours} hours, you should receive an email from us saying that it has been cancelled. This is your cue to <a href="https://www.swapmyvote.uk/?utm_source=email&utm_medium=welcome_email&utm_campaign=ge2019">choose a new voting partner</a>.
+	If the swap is not confirmed within #{swap_validity_hours} hours, you should receive an email from us saying that it has been cancelled. This is your cue to <a href="https://www.swapmyvote.uk/?utm_source=email&utm_medium=welcome_email&utm_campaign=ge2019">choose a new voting partner</a>. However we want to find as many people as possible a good swap, so <strong>for the final few days of the campaign we will be shortening the time for people to confirm a swap</strong>.
 %p
 	If we all tell the people we know, it won’t be long before there’s a critical mass working together to make votes count for more. As soon as we can pair you up, we will email you on this address with the details of your new voting partner.
 %p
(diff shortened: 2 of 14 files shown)


app/models/user.rb
```diff
@@ -24,6 +24,8 @@ class User < ApplicationRecord
   after_save :send_welcome_email, if: :needs_welcome_email?
   before_destroy :clear_swap
 
+  include ApplicationHelper
```
aspiers: This actually bloats `User` quite a bit with stuff it doesn't need - could we perhaps put the helper method under a separate `swaps_helper.rb`?

baob: Yes, I can do that ... make a lot of sense

baob:
> This is great as far as it goes, but it also needs to affect the various web and email view templates which mention the swap expiry duration.

Really .. I did go hunting but didn't see anything related to the keywords swap or 'expir' . Can you point me at a couple of examples, the I'll figure out the keywords and find the rest

aspiers:
[**@baob**](https://github.com/baob) commented on [November 27, 2019 11:15 PM](https://github.com/swapmyvote/swapmyvote/pull/180#issuecomment-559282212 "2019-11-27T23:15:52Z - Replied by Github Reply Comments"):
> Really .. I did go hunting but didn't see anything related to the keywords swap or 'expir' . Can you point me at a couple of examples, then I'll figure out the keywords and find the rest

"hours" is the magic word you were looking for ;-)  Although as you probably already figured out, expiry of confirmed swaps is a totally different thing to expiry of *potential* swaps.

You'll also find a mention in the FAQ.

```
$ git grep hours                                                                                                                                           
app/models/user.rb:    # Clear out swaps every few hours to keep the list fresh for people checking back
app/models/user.rb:    potential_swaps.where(["created_at < ?", DateTime.now - 2.hours]).destroy_all
app/views/static_pages/faq.html.haml:    %p If your partner hasn't confirmed the swap, it will now automatically cancel after 48 hours and we will email both >
app/views/static_pages/faq.html.haml:        waiting for your proposed partner to confirm a swap - they only have <strike>48</strike> - <strike>24</strike> - >
app/views/user/swaps/_list_potential_swaps.html.haml:   This list will refresh in 2 hours if none of these are a good match for you.
app/views/user_mailer/confirm_swap.html.haml:   If we don't get a confirmation in the next 6 hours, we'll have to look for someone else to swap
app/views/user_mailer/welcome_email.html.haml:  We want to find as many people as possible a good swap, so, for the final few days of the campaign, we are sho>
app/views/users/show/_confirm_outgoing_swap.html.haml:  If we don't hear back from them in 6 hours, we'll cancel the swap and you can pick someone else.
lib/tasks/swaps.rake:  desc "Cancel swaps which are older than 24 hours"
lib/tasks/swaps.rake:    swaps = Swap.where({confirmed: false}).where(['created_at < ?', DateTime.now - 6.hours])
```

app/models/user.rb
```diff
@@ -34,7 +36,7 @@ def omniauth_tokens(auth)
 
   def potential_swap_users(number = 5)
     # Clear out swaps every few hours to keep the list fresh for people checking back
-    potential_swaps.where(["created_at < ?", DateTime.now - 2.hours]).destroy_all
+    potential_swaps.where(["created_at < ?", DateTime.now - swap_validity_period]).destroy_all
```
aspiers: Ah no sorry, this should not be touched.  Expiry of confirmed swaps is what we want to parameterise, and that's a totally different thing to expiry of *potential* swaps.

baob: > a totally different thing to expiry of potential swaps.

I obviously haven't got my head round what the app does with all the data yet, apologies. Will fix obviously.

baob: Fix pushed

app/models/user.rb
baob: fix pushed

aspiers:
OK, I guess you're still working on the templates and FAQ?  This is definitely M0 BTW.

baob:
@aspiers @tdg We need to come up with an alternative in the template that says:
`so, for the final few days of the campaign, we are shortening the time for people to confirm a swap ...'

Ideas ?

tdg:
https://github.com/swapmyvote/swapmyvote/pull/180#issuecomment-559493003

 The context is that, for previous elections, we had longer expiry times on swap confirmation – I think we may even have started without an expiry time! This election I propose we start with 48 hours. 

The line you mention will come into play a few days before the election when we change to 24 hours and 12 and even 6 - does that answer your question?

baob:
> The line you mention will come into play a few days before the election when we change to 24 hours and 12 and even 6 - does that answer your question?

@tdg I understand the context. What I don't know is what you want to replace this with right now:

"We want to find as many people as possible a good swap, so, for the final few days of the campaign, we are shortening the time for people to confirm a swap, down to just <em>*#{swap_validity_hours} hours *</em>. If the swap is not confirmed within #{swap_validity_hours} hours, you should receive an email from us saying that it has been cancelled. This is your cue to <a href="https://www.swapmyvote.uk/?utm_source=email&utm_medium=welcome_email&utm_campaign=ge2019">choose a new voting partner</a>".

I mean ... I could come up with text if you have no strong feelings, but I thought you might have strong ideas about the text ? Should I just delete the first sentence right now ?

app/views/user_mailer/welcome_email.html.haml
```diff
@@ -19,7 +19,7 @@
 		%li If need be, check that there is a candidate standing there (there is a link below)
 		%li <em>Click your preferred voting partner to propose a vote swap!</em>
 
-	We want to find as many people as possible a good swap, so, for the final few days of the campaign, we are shortening the time for people to confirm a swap, down to just <em>*6 hours*</em>. If the swap is not confirmed within 6 hours, you should receive an email from us saying that it has been cancelled. This is your cue to <a href="https://www.swapmyvote.uk/?utm_source=email&utm_medium=welcome_email&utm_campaign=ge2019">choose a new voting partner</a>.
+	We want to find as many people as possible a good swap, so, for the final few days of the campaign, we are shortening the time for people to confirm a swap, down to just <em>*#{swap_validity_hours} hours *</em>. If the swap is not confirmed within #{swap_validity_hours} hours, you should receive an email from us saying that it has been cancelled. This is your cue to <a href="https://www.swapmyvote.uk/?utm_source=email&utm_medium=welcome_email&utm_campaign=ge2019">choose a new voting partner</a>.
```
aspiers:
```suggestion
	If the swap is not confirmed within #{swap_validity_hours} hours, you should receive an email from us saying that it has been cancelled. This is your cue to <a href="https://www.swapmyvote.uk/?utm_source=email&utm_medium=welcome_email&utm_campaign=ge2019">choose a new voting partner</a>. However we want to find as many people as possible a good swap, so <strong>for the final few days of the campaign we will be shortening the time for people to confirm a swap</strong>.
```

aspiers:
@baob 
[**@baob**](https://github.com/baob) commented on [November 28, 2019 1:55 PM](https://github.com/swapmyvote/swapmyvote/pull/180#issuecomment-559503414 "2019-11-28T13:55:20Z - Replied by Github Reply Comments"):
> I mean ... I could come up with text if you have no strong feelings, but I thought you might have strong ideas about the text ? Should I just delete the first sentence right now ?

I've suggested new text above.

Issue #130, opened 2019-11-23
Require phone verification before confirming a swap


aspiers:
Currently phone verification is entirely optional and has no effect.  We should require it in order to prevent [Sybil attacks](https://en.wikipedia.org/wiki/Sybil_attack).

The right place to enforce the requirement is just as the user has decided a swap partner and is clicking to confirm the swap with that partner, since by then they should be sufficiently invested to be willing to do a tiny bit of extra work.

N.B. We also want to allow verification earlier - it's currently possible via the user's profile page, and via #129 should also become possible just after first login - but this issue is only about the *hard* requirement at the time of swap confirmation.

tdg:
I _think_ the right time to ask is when setting constituency and (for Twitter logins) email. Slightly difficult to check this instinct without the Twitter login path (currently broken)

aspiers:
This one is **SUPER** important.  Any takers?  @threedaymonk or @egglestn or @fkorning  maybe?

fkorning:
How's this: we add 2 fields to mobile_phones. (1) locale cc code, and (2) a randomly generated base 62 OTP activation token.  an activation SMS message is then sent to the mobile number.  it includes a url with the activation code (say ?code=Ghxy17Jfwk90).

fkorning:
or is the activation code already there?  what's the verify_id field? (create_mobile_phones.rb)

tdg:
We already have verifiable phone number (cf https://github.com/swapmyvote/swapmyvote/issues?q=mobile+label%3A%22mobile+verification%22 ) 

- This issue is about making verification mandatory at a **relevant moment** in the user journey.

The latest it could happen is when someone proposes or confirms a swap, respectively. 
The earliest it could happen would be at the same time the user specifies their constituency.

aspiers:
Right, this issue is about the latest time. There is already a separate
issue filed about the earliest time.

aspiers:
Fixed by #187.

PR #187, opened 2019-11-28
Enforce phone verification

egglestn:
Don't allow users to accept or suggest a swap until they have verified their phone number

New tests for the swaps controller

I *think* I have caught all the places where hound was complaining about line length...

Screen shots:
image

image

### DIFF
--- a/app/controllers/user/swaps_controller.rb
+++ b/app/controllers/user/swaps_controller.rb
@@ -29,7 +29,6 @@ class User::SwapsController < ApplicationController
   end
 
   def update
-    update_phone
     if swap_params[:confirmed] == "true"
       @user.confirm_swap
     end
@@ -60,19 +59,6 @@ class User::SwapsController < ApplicationController
   def assert_parties_exist
     return if @user.willing_party && @user.preferred_party
     redirect_to edit_user_path
-  end
-
-  def update_phone
-    return if phone_param.blank?
-    return if @user.mobile_number == phone_param
-
-    begin
-      @user.mobile_number = phone_param
-    rescue ActiveRecord::RecordInvalid
-      flash[:errors] = @user.mobile_phone.errors.full_messages
-      redirect_to edit_user_path
-      return
-    end
   end
 
   def swap_params
--- a/app/controllers/user/swaps_controller.rb
+++ b/app/controllers/user/swaps_controller.rb
@@ -3,7 +3,7 @@ class User::SwapsController < ApplicationController
   before_action :require_login
   before_action :assert_incoming_swap_exists, only: [:update, :destroy]
   before_action :assert_parties_exist, only: [:show]
-  before_action :assert_mobile_phone_verified, only: [:new, :create]
+  before_action :assert_mobile_phone_verified, only: [:new, :create, :update]
 
   include UsersHelper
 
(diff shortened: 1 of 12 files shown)


app/controllers/users_controller.rb
```diff
@@ -49,8 +54,8 @@ def user_params
   end
 
   def phone_param
-    # We get the value from mobile_phone[full] on the form, intl-tel-input
-    # will put a normalised version of the number (with IDD code) there for us.
+    return params[:mobile_phone][:number] if params[:mobile_phone] && params[:mobile_phone][:number]
```
egglestn: I had a fight with haml and haml won, so the controller might get different parameters depending on where the (same) partial was linked from. Happy to fix if someone can provide the magic incantation to fix the way the partial is wrapped. Alternatively, we could not use the partial from the user/edit page, and have a different method to update only the phone number

app/controllers/user/swaps_controller.rb
```diff
@@ -38,6 +41,13 @@ def destroy
 
   private
```
egglestn: We should not get here normally, because the UI has some filters and won't show you swaps unless your mobile phone number has been verified. Just in case

jpallen:
Tested this and it all seems to work as expected. I didn't see the screenshot in the original comment - instead I got directed to the profile page, but it seemed to work and flow ok still.

I did hit an issue with needing to make sure I had a leading zero in my phone number, and no error message when I didn't - I just didn't get any code sent. I think that's a separate issue though, and not in scope to fix here.

aspiers:
[**@jpallen**](https://github.com/jpallen) commented on [November 29, 2019 12:32 PM](https://github.com/swapmyvote/swapmyvote/pull/187#issuecomment-559775698 "2019-11-29T12:32:14Z - Replied by Github Reply Comments"):
> Tested this and it all seems to work as expected. I didn't see the screenshot in the original comment - instead I got directed to the profile page, but it seemed to work and flow ok still.

There were two screenshots.  I totally changed the first one because I thought it was the wrong place in the flow to introduce friction.  However while the second screenshot is out of date (I fixed the centering and mobile phone field), it is essentially still there with the same structure.  Can you double-check you saw that one and it worked?

app/views/mobile_phone/_form.html.haml
```diff
@@ -0,0 +1,12 @@
+%p
+  My phone number is
+  = telephone_field_tag "mobile_phone[number]", mobile_number, size: 14
```
stevebaxter: I think we need to use `mobile_phone[full]` here so we get and set the data from the hidden input that is managed by the phone validation widget. This should fix/prevent #209.

aspiers: You can use GitHub's suggestion feature for little tweaks like this :-)  Look:

```suggestion
  = telephone_field_tag "mobile_phone[full]", mobile_number, size: 14
```

aspiers: Hmm, except GitHub is too stupid to allow me to commit that via the web UI, even though I can force-push to this branch :-(

aspiers:
OK this works well for me now.  Could be prettier but hey!

aspiers:
This fixes #130.

PR #194, opened 2019-11-28
Dont show zero vote polls

baob:
This was something I worked on a day or two ago ... discussed with @tdg ... where a particular party has 0 'votes' in a constituency, just hide that column.


baob:
Example of how it looks after change
image

Versus before
image

PR #207, opened 2019-11-29
Allow Facebook log in to be disabled

### LINKED ISSUE #201: Disable log ins with Facebook
Until Facebook reviews and approves our app, we need to disable Facebook log ins to be able to put the site live. If this can be done with a config variable like `DISABLE_FACEBOOK_LOGIN` that can easily be switch, that would be great, but something simple and quick is most important.

jpallen:
Adds a `DISABLE_LOG_INS` env var, which if set to `facebook` will disable Facebook logins (via the `log_in_with` helpers, and show a message about it coming soon.

I'm changed `log_in_method_restricted` -> `log_in_method_unrestricted` (note the `un-`), because the semantics of this seemed to be backwards with what it actually returned. I might have missed something though.

### Screenshot with Facebook disabled

Screenshot 2019-11-29 at 10 54 26

Closes #201.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -58,16 +58,19 @@ module ApplicationHelper
   # will be shown but the dialog is not shown automatically.
 
   def log_in_with_facebook?
+    return false if %w[all facebook].include? ENV["DISABLE_LOG_INS"]
     return true unless params[:log_in_with]
     return %w[any facebook].include? params[:log_in_with]
   end
 
   def log_in_with_twitter?
+    return false if %w[all twitter].include? ENV["DISABLE_LOG_INS"]
     return true unless params[:log_in_with]
     return %w[any twitter].include? params[:log_in_with]
   end
 
-  def log_in_method_restricted?
+  def log_in_method_unrestricted?
+    return false if ENV["DISABLE_LOG_INS"]
     return !params[:log_in_with] || params[:log_in_with] == "any"
   end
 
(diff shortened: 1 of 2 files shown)


app/helpers/application_helper.rb
```diff
@@ -58,16 +58,19 @@ def mobile_needs_verification?
   # will be shown but the dialog is not shown automatically.
 
   def log_in_with_facebook?
+    return false if %w[all facebook].include? ENV["DISABLE_LOG_INS"]
     return true unless params[:log_in_with]
     return %w[any facebook].include? params[:log_in_with]
   end
 
   def log_in_with_twitter?
+    return false if %w[all twitter].include? ENV["DISABLE_LOG_INS"]
```
aspiers: Why disable Twitter if `ENV["DISABLE_LOG_INS"] == "facebook"`?

app/helpers/application_helper.rb
```diff
@@ -58,16 +58,19 @@ def mobile_needs_verification?
...
     return true unless params[:log_in_with]
     return %w[any twitter].include? params[:log_in_with]
   end
 
-  def log_in_method_restricted?
+  def log_in_method_unrestricted?
```
aspiers: Ah yes, you're right it was named the wrong way around.

app/helpers/application_helper.rb
jpallen: it's not, this line will disable twitter if `ENV["DISABLE_LOG_INS"] == "twitter"`

aspiers: OK you can tell I haven't had enough sleep recently ... :man_facepalming:

PR #213, opened 2019-11-29
configure haml_lint and guard-haml_lint (#141)

### LINKED ISSUE #141: add HAML linting
We need HAML linting locally, just like we have rubocop and scss_lint.  @aspiers already has the changes for this, but it also requires a ton of fixes for HAML lint.

- https://github.com/sds/haml-lint/blob/master/lib/haml_lint/linter/README.md

aspiers:
Identical setup to scss_lint.

Also add a new 'lint' task to Rake which runs all three linters.

Will fix #141 when all the linting issues are addressed.


aspiers:
Posting this before it gets lost in my local repo, but I recommend we avoid this until M4 because it will cause merge conflicts with other PRs and slow down development.

stevebaxter (dismissed):
LGTM from a JS POV, not much change there though!

baob:
There's an updated branch at https://github.com/baob/swapmyvote/pull/new/haml-lint  but it produces an awful lot of warnings ... and I suspect they will need manual changes, not automated corrections like rubocop.

aspiers:
@baob Take a look at this :-)

aspiers:
Rebased

PR #214, opened 2019-11-29
Refactor social sharing buttons

aspiers:
Eliminate duplication, make slight improvements to the buttons in
the process, and then add the buttons to the view where user is waiting for their outgoing swap to be confirmed.

### DIFF
--- a/app/views/user/share/show.html.haml
+++ b/app/views/user/share/show.html.haml
@@ -5,7 +5,7 @@
       Thanks! We're looking for someone for you to swap with.
 
     %p.text-center
-      Would you like to share SwapmyVote before we continue?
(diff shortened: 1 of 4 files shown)


app/views/user/share/show.html.haml
```diff
@@ -3,15 +3,12 @@
     %p.searching-for-swap.text-center
       %i.fa.fa-2x.fa-search
       Thanks! We're looking for someone for you to swap with.
+
     %p.text-center
       Would you like to share SwapmyVote before we continue?
```
tdg: Need to capitalise ‘My’ :)

aspiers: Haha, touchee ;-)

PR #215, opened 2019-11-30
Add notion of test users

aspiers:
Support skipping mobile verification for test users.  This is useful for testing, and also required to pass Facebook's app review since they won't want to enter mobile numbers when testing the use of `user_link`.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -74,9 +74,9 @@ class User < ApplicationRecord
     PotentialSwap.destroy(incoming_potential_swaps.pluck(:id))
   end
 
-  def phone_verified?
-    return false unless mobile_phone
-    mobile_phone.verified
+  # This allows mocking in tests
+  def mobile_phone_verified?
+    mobile_phone&.verified
   end
 
   def swap_with_user_id(user_id)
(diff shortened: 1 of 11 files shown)


aspiers:
@baob Any chance you could review this? Felt slightly rushed so wouldn't mind a 2nd pair of eyes.  You'll find it much easier to review commit by commit, oldest first.

app/models/user.rb
```diff
@@ -155,8 +155,12 @@ def send_vote_reminder_email
     UserMailer.reminder_to_vote(self).deliver_now
   end
 
+  def name
+    self[:name].try { |n| n + test_user_suffix }
+  end
```
baob: I feel a bit uncomfortable about interfering with what ActiveRecord says about the name ... but I can see why you did it. The alternative would mean choosing another method name to represent the concept (`name_for_ui` perhaps ? ... a bit yuk) and then changing callers where appropriate. maybe later, on a refactoring ?

spec/models/user_spec.rb
```diff
@@ -245,4 +245,67 @@
...
+      end
+
+      it "example.com email returns true" do
+        subject.email = "foo@example.com"
+        expect(subject.mobile_verification_missing?).to be_truthy
+      end
+
+      context "when skipping test user verification," do
+        before do
+          ENV["TEST_USERS_SKIP_MOBILE_VERIFICATION"] = "1"
```
baob: I do this setup with mocks ... something like
```
allow(ENV). to receive(:[]).with("TEST_USERS_SKIP_MOBILE_VERIFICATION").and_return("1")
```
nothing to clean up then.

app/models/user.rb
aspiers: Agreed on all points! I dunno haha

spec/models/user_spec.rb
aspiers: Doh yeah that's nicer. No time to fix now :-(

Issue #216, opened 2019-11-30
Record explicit permission to process personal data


stevebaxter:
When a user creates an account, we must request their consent to store, process and share their personal data, including their political opinions.

It may be enough to show a checkbox at the bottom of the "sign in choices" screen along the lines of:

---
☐ By signing up, I consent to SwapMyVote processing my personal data, including my voting preference, and sharing this data with my potential swap by email or text message

We will never share anything with anyone else about your voting preferences or post to your feed without your permission.

Why do I need to log in with Facebook or Twitter?

---

The checkbox should be unchecked by default, the user shouldn't be allowed to proceed until it is checked. We will record the timestamp of the agreement as the time the record was created in the database (this is already present).

We should probably also link directly to our privacy policy and more information about what we are processing and how from here.

PR #237, opened 2019-12-01
Add consent checkbox to sign-up/in screen

stevebaxter:
This is pretty quick and dirty, we only do this client-side and we use HTML5 form validation to do the work. It should be sufficient to get us up and running though.

Also fixed an issue in the FAQ I noticed when testing this.

fixes: #216

### DIFF
--- a/app/views/layouts/_login.html.haml
+++ b/app/views/layouts/_login.html.haml
@@ -24,8 +24,9 @@
         working on adding support for signing up with an email address. In
         the meantime, you can log in with Twitter.
     %p.text-center
-      %input#data-consent{type: "checkbox", name: "data-consent", required: true}
-      %label{for: "data-consent"}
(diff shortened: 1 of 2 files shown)


app/views/layouts/_login.html.haml
```diff
@@ -23,8 +23,17 @@
         We're still waiting for our Facebook app to be approved, and we're
         working on adding support for signing up with an email address. In
         the meantime, you can log in with Twitter.
+    %p.text-center
+      %input#data-consent{type: "checkbox", name: "data-consent", required: true}
+      %label{for: "data-consent"}
+        I consent to SwapMyVote processing my personal data
+    %p.text-center.small.subdued
+      We will share your voting preferences and the public link to your
```
stevebaxter: Text reviewed by @tdg

PR #247, opened 2019-12-02
Improve UX of postcode input

coliff:
- stops annoying squiggly line for spelling mistakes in postcodes
- setting maxlength is a good idea

(you could optionally also add minlength: 5


aspiers:
Great idea! Please can you add minlength then?

PR #263, opened 2019-12-03
Cast image urls to protocol agnostic URLs

### LINKED ISSUE #259: Facebook profile photos not showing
The Facebook profile photos aren't  showing, it looks like we have an http: link when it should be https:

>Mixed Content: The page at 'https://www.swapmyvote.uk/user/edit' was loaded over HTTPS, but requested an insecure image 'http://graph.facebook.com/v3.0/10152709497637237/picture'. This content should also be served over HTTPS.

jpallen:
Fixes #259

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -196,7 +196,7 @@ class User < ApplicationRecord
   end
 
   def image_url
-    identity&.image_url
+    identity&.image_url&.gsub('http://', '//')
   end
 
   def provider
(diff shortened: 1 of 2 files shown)


stevebaxter:
Now I see a different problem:

>Failed to load resource: net::ERR_BLOCKED_BY_CLIENT

jpallen:
@stevebaxter Ad-blocker maybe?

stevebaxter:
Yes it was, works perfectly now.

PR #268, opened 2019-12-04
Cancel swap no user

### LINKED ISSUE #266: [Airbrake] [Production] undefined method `email' for nil:NilClass
**Airbrake error:** [#7995](https://herokuapp32817511herokucom.airbrake.io/projects/110137/groups/2626851525378677995)
**Airbrake project:** swapmyvote

**Error type:** `NoMethodError`
**Error message:** `` undefined method `email' for nil:NilClass ``
**Where:** `users#update`
**Occurred at:** Dec 04, 2019 08:45:03 UTC
**First seen at:** Dec 04, 2019 08:45:03 UTC
**Occurrences:** 0 (0 since last deploy on Dec 03, 2019 21:12:34 UTC)
**Severity:** `error`

**URL:** [https://www.swapmyvote.uk/user](https://www.swapmyvote.uk/user)
**File:** `/PROJECT_ROOT/app/mailers/user_mailer.rb`

**Backtrace:**
```
/PROJECT_ROOT/app/mailers/user_mailer.rb:28:in swap_cancelled
/GEM_ROOT/gems/actionpack-5.2.3/lib/abstract_controller/base.rb:194:in process_action
/GEM_ROOT/gems/actionpack-5.2.3/lib/abstract_controller/callbacks.rb:42:in block in process_action
/GEM_ROOT/gems/activesupport-5.2.3/lib/active_support/callbacks.rb:98:in run_callbacks
```

egglestn:
Cancel swap was falling over in a heap if the user had gone away (~~I assume closed their account,~~ there may be other ways to get here **UPDATE: but we don't know how as per #288), resulting in airbrake errors.

Add tests for the cancel swap mailer
Tolerate null values for both ends of the swap.

Closes #266.

### DIFF
--- a/app/mailers/user_mailer.rb
+++ b/app/mailers/user_mailer.rb
@@ -25,10 +25,13 @@ class UserMailer < ApplicationMailer
   end
 
   def swap_cancelled(user, swap_with)
-    return nil if user.email.blank?
+    return nil if user&.email.blank?
     @user = user
     @swap_with = swap_with
-    mail(to: @user.email, subject: "Your swapped vote with #{swap_with.name} has been cancelled.")
+    mail(to: @user.email,
+         subject: swap_with ?
+           "Your swapped vote with #{swap_with.name} has been cancelled."
+         : "Your swapped vote has been cancelled.")
   end
 
   def not_swapped_follow_up(user)
--- a/app/views/user_mailer/swap_cancelled.haml
+++ b/app/views/user_mailer/swap_cancelled.haml
@@ -1,21 +1,21 @@
 %p
-	Hi #{@user.name},
+  Hi #{@user.name},
 %p
-	I'm sorry to say that your swap with #{@swap_with.name} has been cancelled.
-	This could have happened if you or #{@swap_with.name} changed preferred parties, moved constituency,
-	declined the swap, or (most likely!) if the swap expired before being confirmed.
+  I'm sorry to say that your swap with #{@swap_with.name} has been cancelled.
+  This could have happened if you or #{@swap_with.name} changed preferred parties, moved constituency,
+  declined the swap, or (most likely!) if the swap expired before being confirmed.
 %p
-	= link_to "Find another vote swapping partner - log in to swapmyvote.uk now.", user_url(log_in_with: @user.provider)
-	Don't delay!
+  = link_to "Find another vote swapping partner - log in to swapmyvote.uk now.", user_url(log_in_with: @user.provider)
+  Don't delay!
 %p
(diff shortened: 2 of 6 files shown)


app/models/user.rb
```diff
@@ -196,7 +196,7 @@ def mobile_verification_missing?
   end
 
   def image_url
-    identity&.image_url&.gsub('http://', '//')
+    identity&.image_url&.gsub("http://", "//")
```
egglestn: Guard -> rubocop complained about this, no idea what changed because it isn't new code

config/environments/test.rb
```diff
@@ -43,6 +43,9 @@
 
   # Raises error for missing translations
   # config.action_view.raise_on_missing_translations = true
+
+  # Default mailer URL for test env only!
+  config.action_mailer.default_url_options = { host: "www.example.com" }
```
egglestn: Possibly overkill for this specific bugfix, but could come in handy for improving the mailer tests in future

aspiers: Good idea

app/mailers/user_mailer.rb
```diff
@@ -25,10 +25,10 @@ def swap_confirmed(user, swap_with)
   end
 
   def swap_cancelled(user, swap_with)
-    return nil if user.email.blank?
+    return nil if user&.email.blank?
     @user = user
     @swap_with = swap_with
-    mail(to: @user.email, subject: "Your swapped vote with #{swap_with.name} has been cancelled.")
+    mail(to: @user.email, subject: "Your swapped vote with #{swap_with&.name} has been cancelled.")
```
aspiers: If we don't know the name we have to change the grammar - I've done this.

app/models/user.rb
aspiers: Actually no it was a bug in `.travis.yml` which I've fixed.

app/views/user_mailer/swap_cancelled.haml
```diff
@@ -1,9 +1,10 @@
 %p
 	Hi #{@user.name},
 %p
-	I'm sorry to say that your swap with #{@swap_with.name} has been cancelled.
-	This could have happened if you or #{@swap_with.name} changed preferred parties, moved constituency,
-	declined the swap, or (most likely!) if the swap expired before being confirmed.
+	I'm sorry to say that your swap with #{@swap_with&.name} has been cancelled.
```
aspiers: Again this will interpolate `nil` into the text which we don't want - I'll make the mention of the name conditional.

Issue #70, opened 2019-11-09
Prevent swaps with same party


aspiers:
Currently you can select the same party for your preferred party *and* the one you're willing to vote for, both when logged in, and on the pre-login home screen.  We need some Javascript validation to prevent both, but also ideally some backend validation for the logged in case too.

In the logged in scenario, this is used as a "feature", as a way of allowing a user to indicate a desire to "deactivate" their account - explained in the FAQ at `/faq#deactivate`.  However we should replace that with a much better way of properly removing user accounts.

DouglasLutz:
I can work on this

aspiers:
Thanks Douglas, any help is hugely appreciated!  Although we do have
significantly more urgent issues on the project board right now if
you're not particularly attached to this one.

DouglasLutz:
Sure @aspiers, I will try to help with other ones too

aspiers:
I think this is much more urgent now there are only two constituencies.

aspiers:
I've filed #466 for the pausing feature, so that we can address the immediate usability issue for BE2021 and not get bogged down with the functionality regression.

PR #279, opened 2019-12-05
Swap prevention with same party

DouglasLutz:
# Solves #70 

## Description:
Added validation to user to check if it has selected the same preferred party and willing party.
The steps where:
* Create the user model validator `DistinctPartiesValidator`
* Add tests to the validation
* Add front-end validation in the home page when user is not signed in

## Notes: 
I did not add front-end validation when the user is signed in because it already shows the error message with `flash[:errors]`

## How it was tested?
Running rspec and rubocop

## Screenshots
Back-end validation
parties_validation

Front-end validation
js-validation

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -23,6 +23,7 @@ class UsersController < ApplicationController
 
   def update
     @user.update(user_params) if params[:user]
+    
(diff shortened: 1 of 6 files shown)


aspiers:
@cla-bot check

aspiers:
Superseded by #455.

PR #287, opened 2019-12-06
Issue #277 Offer better swaps

### LINKED ISSUE #277: Offer better swaps
@baob is working on an idea involving surfacing swaps in more marginal willing.partner preferred.user places

baob:
Closes #277 

--------------------

For each combination of party and constituency we figure out how close that party is to tipping the balance in that constituency.

Method: Take the party of interest, and find their predicted (by EC) vote share in that constituency.
Then take the rest of the parties in that constituency and find the maximum vote share amongst them.
Then find the difference and save it as an absolute value. call it marginal score or something.

Now we can offer voters an attractive constituency to swap with, by preferring constituencies where their preferred party has a low marginal score ... meaning the party either needs a bit of help getting over the line, or it needs its expected majority bolstering to safe levels. I’m proposing that when we offer potential swaps, 50% of those will be random, and 50% will be from marginals as defined above.

big marginal score implies the preferred party is either way out in front and a safe seat, or way behind without a chance

### DIFF
--- a/app/models/ons_constituency.rb
+++ b/app/models/ons_constituency.rb
@@ -6,4 +6,8 @@ class OnsConstituency < ApplicationRecord
   has_many :recommendations,
            primary_key: "ons_id",
            foreign_key: "constituency_ons_id"
+
+  def parties_by_marginal_score
+    polls.order(:marginal_score).map(&:party)
+  end
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -59,7 +59,9 @@ class User < ApplicationRecord
   end
 
   def try_to_create_marginal_swap
-    swaps = complementary_voters.where({ constituency_ons_id: marginal_polls.map(&:constituency_ons_id) })
+    swaps = complementary_voters.where(
+      { constituency_ons_id: marginal_polls.map(&:constituency_ons_id) }
+    )
     return one_swap_from_possible_users(swaps)
   end
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -106,7 +106,7 @@ class User < ApplicationRecord
 
   private def one_swap_from_possible_users(user_query)
     offset = rand(user_query.count)
-    target_user = user_query.offset(offset).limit(1).first
+    target_user = user_query.offset(offset).take
     return nil unless target_user
     # We need emails to send confirmation emails
     return nil if target_user.email.blank?
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -121,7 +121,7 @@ class User < ApplicationRecord
   end
 
   def marginal_polls
-    Poll.where(["marginal_score < ?", 1000]).where(party: preferred_party).order(:marginal_score)
+    Poll.where(["marginal_score < ?", 1000]).where(party: preferred_party)
   end
 
   def marginal_constituencies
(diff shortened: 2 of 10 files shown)


app/models/user.rb
```diff
@@ -45,19 +45,36 @@ def potential_swap_users(number = 5)
   def create_potential_swaps(number = 5)
     max_attempts = number * 2
     while potential_swaps.reload.count < number
+      try_to_create_marginal_swap
+      max_attempts -= 1
```
aspiers: So this uses up attempts at twice the speed - the question is whether it compensates for that equally by producing swap candidates at twice the speed.  I suspect it's lower than twice because it's presumably much harder to find a marginal swap than a normal potential swap.  So this might result in fewer potential swaps being generated than before, or am I misunderstanding?

app/models/ons_constituency.rb
```diff
@@ -6,4 +6,8 @@ class OnsConstituency < ApplicationRecord
   has_many :recommendations,
            primary_key: "ons_id",
            foreign_key: "constituency_ons_id"
+
+  def parties_by_marginal_score
```
aspiers: Where is this used?  And is this ordered by score descending or ascending?

app/models/user.rb
```diff
@@ -45,19 +45,36 @@ def potential_swap_users(number = 5)
...
       willing_party_id: preferred_party_id
-    ).where("constituency_ons_id like '_%'")
-    offset = rand(swaps.count)
-    target_user = swaps.offset(offset).limit(1).first
+    )
+  end
+
+  private def one_swap_from_possible_users(user_query)
+    offset = rand(user_query.count)
+    target_user = user_query.offset(offset).limit(1).first
```
aspiers:
```suggestion
    target_user = user_query.offset(offset).take
```

although I haven't tested that's correct.

app/models/user.rb
```diff
@@ -71,6 +88,14 @@ def try_to_create_potential_swap
     return potential_swaps.create(target_user: target_user)
   end
 
+  def marginal_polls
+    Poll.where(["marginal_score < ?", 1000]).where(party: preferred_party).order(:marginal_score)
```
aspiers: Aren't we randomly sampling from this list?  In which case the ordering is pointless?  Or am I missing something?

app/models/user.rb
baob: I think you're right ... I think it makes sense to either double the max_attempts at start or else decrement once in the loop ... probably the former, as it keeps the name 'max_attempts' a bit more honest.

app/models/ons_constituency.rb
baob: You're right ... not used ... I should apply YAGNI and delete, particularly as it's trivial to recreate

app/models/user.rb
baob: Yes ... in earlier stages of hacking I had an idea about weighting the randomisation towards low marginal scores, rather than applying a cutoff ... plus it was useful in the console. now it should go.

app/models/ons_constituency.rb
baob: I see your comment @aspiers so I will leave alone
> Adam Spiers 10:14 AM
> @baob  That method might be useful when debugging on console
> so I suggest keeping

app/models/user.rb
baob: @aspiers more elegant solution now on the branch

app/models/user.rb
baob: take turns out to be less db load. 👍 change now pushed to branch

app/models/user.rb
baob: Change just pushed

PR #290, opened 2019-12-06
Live from brexit data refresh dec 06

baob:


aspiers:
This works well for me locally but adds the avaaz recos on the far right hand side next to the loonies - is it possible to group with the other Remain recos?

baob:
> This works well for me locally but adds the avaaz recos on the far right hand side next to the loonies - is it possible to group with the other Remain recos?

Not without introducing special cases and value judgements about what's going on.  LFB doesn't classify recommenders as remain or leave. If I make a special case out of avazz and put it first ... but that's still implicitly making a value judgement that what this election is about is brexit and we've decided which side of the line these recommenders are on .

The only thing I could reasonably  do without bias is order them alphabetically ... even ordering them as per the the LFB json would might take a bit of work

tdg:
comparethetacticals.com has avaaz down as ‘stop conservatives’

PR #310, opened 2019-12-08
Issue #277 sort marginals to top

baob:


app/models/user.rb
```diff
@@ -32,14 +32,20 @@ def omniauth_tokens(auth)
...
     # Clear out swaps every few hours to keep the list fresh for people checking back
     potential_swaps.where(["created_at < ?", DateTime.now - 2.hours]).destroy_all
     create_potential_swaps(number)
     swaps = potential_swaps.all.eager_load(
-      target_user: { constituency: [{ polls: :party }] }
+      target_user: [ :identity, { constituency: [{ polls: :party }] } ]
     )
-    return swaps.map {|s| s.target_user}
+    return swaps
+      .sort_by { |s| s.target_user.willing_party_poll&.marginal_score || 999_999 }
```
aspiers: This looks really expensive - one query per sort comparison. Can't you sort in the SQL? or at very least use a Schwartzian transform?

https://en.wikibooks.org/wiki/Algorithm_Implementation/Sorting/Schwartzian_transform#Full_Schwartzian_Transform

If the latter, I guess it's lucky that we only need to sort by one criteria, because there is a bug in the above code which I described here:

https://stackoverflow.com/questions/48240069/idiomatic-lazy-sorting-by-multiple-criteria

baob: I spent several hours trying to get that to work, doesn't seem to be a way do it through active record. You need to join polls twice, once to get all polls for the constituency, and once to get only the poll for the other users 'willing party' (our 'preferred party') in order to sort by marginal score.

It's not expensive on sqllite ... it's much less expensive (by a factor of 30) than the query that retrieves all the 5 potential swaps.

aspiers: Ah OK, `sort_by` has Schwartzian transform built in - I guess this is good enough then.

baob: I was just about going to post the log to demonstrate the sort key calculation happened only once, which is what the transform is all about.
```
  SQL (31.9ms)  SELECT "potential_swaps"."id" AS t0_r0, "potential_swaps"."source_user_id" AS t0_r1,.... WHERE "potential_swaps"."source_user_id" = ?  [["source_user_id", 61]]
  ↳ app/models/user.rb:47
  Poll Load (0.2ms)  SELECT  "polls".* FROM "polls" WHERE "polls"."constituency_ons_id" = ? AND "polls"."party_id" = ? LIMIT ?  [["constituency_ons_id", "E14000964"], ["party_id", 4], ["LIMIT", 1]]
  ↳ app/models/user.rb:36
  Poll Load (0.1ms)  SELECT  "polls".* FROM "polls" WHERE "polls"."constituency_ons_id" = ? AND "polls"."party_id" = ? LIMIT ?  [["constituency_ons_id", "E14000552"], ["party_id", 4], ["LIMIT", 1]]
  ↳ app/models/user.rb:36
  Poll Load (0.1ms)  SELECT  "polls".* FROM "polls" WHERE "polls"."constituency_ons_id" = ? AND "polls"."party_id" = ? LIMIT ?  [["constituency_ons_id", "W07000064"], ["party_id", 4], ["LIMIT", 1]]
  ↳ app/models/user.rb:36
  Poll Load (0.5ms)  SELECT  "polls".* FROM "polls" WHERE "polls"."constituency_ons_id" = ? AND "polls"."party_id" = ? LIMIT ?  [["constituency_ons_id", "E14000669"], ["party_id", 4], ["LIMIT", 1]]
  ↳ app/models/user.rb:36
  Poll Load (0.2ms)  SELECT  "polls".* FROM "polls" WHERE "polls"."constituency_ons_id" = ? AND "polls"."party_id" = ? LIMIT ?  [["constituency_ons_id", "E14000733"], ["party_id", 4], ["LIMIT", 1]]
  ↳ app/models/user.rb:36
```

Issue #44, opened 2019-11-04
Allow user to be linked with multiple social accounts


aspiers:
Currently login is either facebook or twitter, but it should be possible to link an account to both (and maybe others later?)  For this we'll probably need a couple of new tables mapping a many-to-many relationship between users and social networks.  Or we could just add facebook and twitter fields to the user table - quick and easy although also kind of ugly.

egglestn:
Feels like multiple tables and many-to-many would be more future proof. Do we have a definite decision?

tdg:
One relevant user need is the opportunity to 'reach out' to your partner over social media. We have had a few pairs who were siloed in the other network (ie one was only on FB, one only on Twitter).

aspiers:
[**@egglestn**](https://github.com/egglestn) commented on [November 8, 2019 10:41 AM](https://github.com/swapmyvote/swapmyvote/issues/44#issuecomment-551554436 "2019-11-08T10:41:10Z - Replied by Github Reply Comments"):
> Feels like multiple tables and many-to-many would be more future proof. Do we have a definite decision?

I agree @egglestn and would happily accept a PR with a db migration to this schema as the first step.  It would of course also need to alter the login workflow accordingly, and also the way the social profiles of swap partners are shared.

The second step would be to allow logged in users to link an (the) other social network to their account, which is going to require modifications to `users_controller.rb`.

aspiers:
@egglestn made the good point that it probably makes sense to introduce this new table before fixing #48.

aspiers:
... although that would change this from M2 to M0.  Hmm...

tdg:
In terms of user need this is a relatively unusual case so, itself, definitely M2

But then I don’t know how much the tables question in #48 is something that would be hard to change later…

aspiers:
Maybe I was misleading you a bit :-)  Introducing a new table is the first step required for this, and that's pretty damn easy so could be done for M0 to help do #48 more cleanly.  The second step (updating the UI to allow connection of a second social account post-login) is more complex but could be done later for M2.

tdg:
Sounds like a great plan 👍

PR #323, opened 2019-12-09
Dynamic mobile number verification

egglestn:
If the user updates the phone number in the UI, update it before sending verification

Also spotted a related issue where there was a messy failure if the phone number was duplicate, it will still fail but you should get a slightly more helpful error message now. A better fix should be along in https://github.com/swapmyvote/swapmyvote/issues/44


aspiers:
Looks good but requires careful review and other emergencies currently consuming my attention :-(

Issue #28, opened 2019-10-31
Add identity verification icons for potential swap partners


aspiers:
When presenting the list of swap candidates, we should put little icons beside each candidate showing how they've verified their identity so far, and by implication (maybe later even explicitly) how much trust can be placed in that candidate being a real person.

aspiers:
This cannot be done until #44 is implemented.

tdg:
We can just use tiny fontawesome icons for mobile, email, facebook, and twitter

tdg:
Can we also show it to the user themselves (about themself) - nice little nudge

tdg:
with title="This user has provided an email" on the HTML element providing useful tooltips on hover
title="This user verified their mobile phone number"
title="This user has a valid Facebook account"

jt-nti:
As far as I understand it, a swap candidate cannot accept a swap until they have verified a phone number. They will definitely have a valid twitter or Facebook (when facebook works) account, to have got in the system, so it seems like it's worth drawing attention to. Maybe even listing unverified accounts below people who are ready to swap.

aspiers:
[**@aspiers**](https://github.com/aspiers) commented on [November 28, 2019 7:13 PM](https://github.com/swapmyvote/swapmyvote/issues/28#issuecomment-559529585 "2019-11-28T15:13:24Z - Replied by Github Reply Comments"):
> This cannot be done until #44 is implemented.

This is not true - I was confused ;-)  We already allow users to verify mobile, and logging in via Twitter or Facebook are forms of identity verification.  The only thing which depends on #44 is being able to display *both* Facebook and Twitter icons on a single user profile.

I think this is very easy to implement and worth doing ASAP -> moving to M2.

aspiers:
Closed by #325.

Issue #102, opened 2019-11-21
Allow users to share email address with confirmed partners


tdg:
This covers both the implementation and the proper obtainment of consent, for which previous discussion was captured in #74.

aspiers:
Phew - turns out we already explicitly covered sharing email (and even phone number) with swap partners in the Personal Data section of the privacy policy pre 2019 launch.

So this issue is more about making it possible and explicit in the UI.  Extra attention is needed in the Facebook/Twitter login cases since sharing email is not as expected there.

baob:
I was going to work on this by adding a consent column to the users table, and associated form changes, but it seems from recent changes to policy wording on the website (on email login branch), there's nothing to do here.

aspiers:
That's not right, we need to track consent server side. This is
complicated, had a long chat with Tom about it last night.

baob:
This is my proposal:
https://github.com/swapmyvote/swapmyvote/issues/324

baob:
@tdg @aspiers  with #324 and #337 now closed I think we can say this one is done ?

aspiers:
Agreed!! Stellar work

Issue #15, opened 2018-02-06
Confirm email address for users


tdg:
Confirm email address for users

aspiers:
@tdg On further investigation I think we misunderstood this issue and that it's about confirming the user's email address.  Having said that, currently it only comes from Facebook or Twitter, so at least in theory it should have already been confirmed by one of those two platforms and we shouldn't need to confirm it.

Also, it turns out that in the production environment we use Mailgun rather than sending email directly, so deliverability based on the sender should be fine.

I think we should deal with CTR in a separate issue.

aspiers:
Hrm, I now see it's possible for the user to update their email address.  So we need to confirm the new email address if this is done.

aspiers:
I just logged in via Twitter and it didn't get my email.  Weird.  @jpallen Any ideas?

aspiers:
Updated description again based on better understanding.

tdg:
>Having said that, currently it only comes from Facebook or Twitter, so at least in theory it should have already been confirmed by one of those two platforms and we shouldn't need to confirm it.

The concern is that, because our email comes from a more obscure domain, it could easily all go to junk mail folders. Would a confirm link help voters train their email clients to treat our emails seriously and deliver to inboxes accordingly?

aspiers:
`.uk` isn't really an obscure domain; I suspect that's not an issue.  No, email clients aren't retrained when you click a confirmation link, but at least it tells us the email address is correct.

tdg:
Fair - but one of the last elections we did have some (ill-defined) problems with this (naked `.uk` was younger back in 2015 :) ). 

Hopefully #40 will resolve the problems, but I still feel it's a good idea to reinforce users' expectations that they will receive - and need to respond to - emails from us

tdg:
Suggest this be moved to M0

baob:
How does this relate to #38 ?  Does that have to be finished first ? or will #38 take care of this completely perhaps ?  @aspiers @jpallen

aspiers:
No they are independent of each other, but email logins increase the need for verification.

baob:
Note: not relevant to BE2022 sol long as both the twitter and facebook logins are disabled. If we manage to do #442 (twitter), this becomes important again.

aspiers:
[...]

PR #325, opened 2019-12-09
Verification icons

egglestn:
Verification icons for phone number, email, twitter, fb

image

### DIFF
--- a/app/views/user/swaps/_swap_profile_inner.html.haml
+++ b/app/views/user/swaps/_swap_profile_inner.html.haml
@@ -1,6 +1,17 @@
 %img.profile-img{ src: other_user.image_url }
 .profile-content
   .profile-name= other_user.redacted_name
+
+  .profile-icons
+    - if other_user.mobile_phone_verified?
+      %i.fa.fa-phone{ title: "Phone number verified" }
+    - if other_user.provider == "twitter"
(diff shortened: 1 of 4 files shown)


app/views/user/swaps/_swap_profile_inner.html.haml
```diff
@@ -1,6 +1,17 @@
...
+
+  .profile-icons
+    - if other_user.mobile_phone_verified?
+      %i.fa.fa-phone{ title: "Phone number verified" }
+    - if other_user.provider == "twitter"
+      %i.fa.fa-twitter{ title: "Twitter account verified" }
+    - if other_user.provider == "facebook"
+      %i.fa.fa-facebook{ title: "Facebook account verified" }
+    - if other_user.provider == "devise_email"
+      %i.fa.fa-envelope{ title: "Email verified" }
```
aspiers: We don't verify email yet sadly - that's #15.  And with Twitter and Facebook we also now require email.  The difference is currently that with email logins we provide the address to the swap partner.  Once #102 is resolved we can share emails even with social login.

```suggestion
      %i.fa.fa-envelope{ title: "Email address available" }
```

aspiers:
I did see a weird glitch where it mucked up the bar graph positioning but it vanished and can't reproduce, so I'm assuming it was a one-off...

aspiers:
Closes #28.

Issue #320, opened 2019-12-08
Reject swap modal is transparent and can't be clicked


tdg:
![Screen Shot 2019-12-08 at 19 31 28](https://user-images.githubusercontent.com/95868/70394966-1b5bc300-19f2-11ea-86c9-586fefa475ea.png)

stevebaxter:
I'll fix this as part of #316. It's a similar set of problems.

Issue #316, opened 2019-12-08
Login modal can't be scrolled on mobile devices


stevebaxter:
On some mobile devices the "login" modal will be poorly positioned and isn't scrollable. In particular, the information on privacy and consent may be clipped and inaccessible. We should make sure this can be scrolled into view.

This was reported by one used on the WhatsApp group.

aspiers:
This was previously tracked in #257 for which there was a different type of fix, but I suspect scrapping `position: fixed` is the main fix.

aspiers:
But also, there shouldn't be a big margin above the modal on mobiles.

aspiers:
Closed by #329.  @stevebaxter FWIW GitHub doesn't properly understand the syntax "fixes: #320, #316" and only auto-closed #320 ;-)

PR #329, opened 2019-12-09
Scrollable modals

stevebaxter:
Use Bootstrap modals instead of custom ones. Including the BS JS screws up jquery-ui a bit, fortunately it's only used for the auto-complete dropdown so we work around it.

Some screenshots:

Screenshot 2019-12-09 14 59 12
Screenshot 2019-12-09 14 59 06
Screenshot 2019-12-09 14 58 47
Screenshot 2019-12-09 14 58 37
Screenshot 2019-12-09 15 07 10

fixes: #320, #316


app/assets/javascripts/autocomplete.js
```diff
@@ -1,140 +1,146 @@
...
+      this.element.hide();
+      this._createAutocomplete();
+      this._createShowAllButton();
+    },
+
+    _createAutocomplete: function() {
+      var element = this.element,
+        selected = element.children(":selected"),
+        value = selected.val()
+          ? selected.text()
```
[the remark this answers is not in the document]
stevebaxter: I don't agree with Hound here (and nor does Prettier).

app/assets/javascripts/autocomplete.js
```diff
@@ -1,140 +1,146 @@
...
-          });
-
-        this._on( this.input, {
-          autocompleteselect: function( event, ui ) {
-            ui.item.option.selected = true;
-            this._trigger( "select", event, {
-              item: ui.item.option
-            });
-          },
+$(function() {
```
stevebaxter: Prettier has reformatted this...

app/assets/javascripts/autocomplete.js
```diff
@@ -1,140 +1,146 @@
...
+      var input = this.input,
+        wasOpen = false;
+
+      $("<a>")
+        .attr("tabIndex", -1)
+        .appendTo(this.wrapper)
+        .addClass(
+          "ui-button ui-widget ui-button-icon-only constituency-autocomplete-toggle ui-corner-right"
+        )
+        .append("<span class='ui-button-icon ui-icon ui-icon-triangle-1-s' />")
```
stevebaxter: This is the key change - `$().button()` is implemented by both BS and jQuery-UI, and BS now wins. We add the button using jQuery manually.

app/assets/javascripts/autocomplete.js
stevebaxter: I've commented on the actual change below.

app/views/home/_swaps_open.html.haml
```diff
@@ -39,16 +39,17 @@
...
   }
 
-.smv-modal.js-error-modal{ style: "display: none;" }
-  .smv-modal-dialog
-    %i.fa.fa-times.subdued.modal-close
-    %p.text-center
-      Please choose both your preferred party and your willing party.
-    %p.text-center
-      %a.btn.btn-primary.modal-close{ href: "#" } OK
+.modal#js-error-modal{ tabindex: "-1", role: "dialog" }
```
stevebaxter: Same change in a few places, we use a BS modal which requires some extra content (`.modal-content`).

app/views/static_pages/faq.html.haml
```diff
@@ -40,7 +40,7 @@
 
     %p
 
-      To prevent fake accounts, we have also made verification of a
+      To help prevent fake accounts, we have also made verification of a
```
stevebaxter: @tdg

app/views/users/edit.html.haml
```diff
@@ -51,7 +51,7 @@
           %p.subdued.small
             = render partial: "why_need_email"
 
-            We need your phone number to prevent people creating fake
+            We need your phone number to help prevent people creating fake
```
stevebaxter: @tdg

app/assets/javascripts/autocomplete.js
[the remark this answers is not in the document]
egglestn: I think it would be clearer all on one line, but I bet hound doesn't agree with that either...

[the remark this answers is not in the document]
stevebaxter: Yeah, these days I just tend to let Prettier auto-format everything. We can probably tweak the rules though...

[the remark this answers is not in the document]
aspiers: Yes feel free to submit better rules!

app/views/static_pages/faq.html.haml
aspiers: Thanks!

app/views/users/edit.html.haml
```diff
@@ -52,8 +52,7 @@
 
           %p.subdued.small
 
-            We only need your email to let you know when we've found a
-            swap partner for you, and your phone number to prevent
+            We only need your phone number to prevent
```
aspiers: I changed this text last night ...

app/assets/javascripts/autocomplete.js
aspiers: Thanks for the explanation!  Tiny nitpick, but in the future I'd recommend putting this stuff in comments in the code, as it will never be found again if only buried in the PR (especially if we want to reap the benefit's of GitHub's "Resolve conversation" button).

stevebaxter:
@aspiers - this shouldn't be needed anymore because Bootstrap manages whether the modal is visible or not. What was the symptom?

stevebaxter:
Thanks for fixing it anyway!

aspiers:
Please look at my fix and log_in_with functionality.

Issue #328, opened 2019-12-09
[Airbrake] [Production] Could not find a valid mapping for path "/auth/facebook/callback"


aspiers:
**Airbrake error:** [#6431](https://herokuapp32817511herokucom.airbrake.io/projects/110137/groups/2630647628976856431)
**Airbrake project:** swapmyvote

**Error type:** `RuntimeError`
**Error message:** `` Could not find a valid mapping for path "/auth/facebook/callback" ``
**Where:** `<no information>`
**Occurred at:** Dec 09, 2019 14:27:12 UTC
**First seen at:** Dec 09, 2019 14:27:12 UTC
**Occurrences:** 0 (0 since last deploy on Dec 08, 2019 21:20:32 UTC)
**Severity:** `error`

[...]

PR #333, opened 2019-12-10
Fix crashes when authentication fails

stevebaxter:
We've been seeing crashes reported in Airbrake with `Could not find a valid mapping for path "/auth/xxxxxx/callback"`. This seems to happen when authentication fails, e.g.:

* CSRF detection on the FB (Oauth2) logins because the `omniauth.state` in the session doesn't match the state passed back in the callback
* Failed login, e.g. login using Twitter but the user cancels the login
* Cancelled FB/Twitter app permission, so the user decides not to give our app permission to access their account

The problem seems to be that Devise can't figure out the route in this situation. To fix this we add a custom `on_failure` callback to ensure the right thing gets called.

See https://github.com/plataformatec/devise/issues/2004#issuecomment-7466322 for more information.

fixes: #328

### DIFF
--- a/app/controllers/users/omniauth_callbacks_controller.rb
+++ b/app/controllers/users/omniauth_callbacks_controller.rb
@@ -72,7 +72,7 @@ class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
     return user
   end
 
-  def retry
+  def failure
     flash[:errors] = ["Login failed! Please try again, or try logging in a different way."]
     destroy
   end
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -309,4 +309,17 @@ Devise.setup do |config|
   # When set to false, does not sign a user in automatically after their password is
   # changed. Defaults to true, so a user is signed in automatically after changing a password.
   # config.sign_in_after_change_password = true
+
+  # Errors from authentication may crash because no Devise mapping is defined
+  # for the callback path. To work around this we manually set the failure
+  # callback to Users::OmniauthCallbacksController.failure().
+  #
+  # See https://github.com/plataformatec/devise/issues/2004#issuecomment-7466322
+  # for more information.
+  OmniAuth.config.on_failure = Proc.new do |env|
+    env['devise.mapping'] = Devise.mappings[:user]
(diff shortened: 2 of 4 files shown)


config/initializers/devise.rb
```diff
@@ -309,4 +309,17 @@
...
   # changed. Defaults to true, so a user is signed in automatically after changing a password.
   # config.sign_in_after_change_password = true
+
+  # Errors from authentication may crash because no Devise mapping is defined
+  # for the callback path. To work around this we manually set the failure
+  # callback to Users::OmniauthCallbacksController.failure().
+  #
+  # See https://github.com/plataformatec/devise/issues/2004#issuecomment-7466322
+  # for more information.
+  OmniAuth.config.on_failure = Proc.new do |env|
```
stevebaxter: This really might not be the right place to put this, however this does seem to work!

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -72,7 +72,7 @@ def create_user_and_identity(auth)
     return user
   end
 
-  def retry
+  def failure
```
stevebaxter: The Devise stuff expects this to be failure().

stevebaxter:
I'm still worried about the root cause of the CSRF issues. This will prevent these causing a crash, but why are they happening at all?

This basically happens here:

https://github.com/omniauth/omniauth-oauth2/blob/master/lib/omniauth/strategies/oauth2.rb#L71

If the state passed back as a URL query param doesn't match `session["omniauth.state"]`, Omniauth rejects the callback. I'm guessing our session has been wiped out somehow, any ideas on how this could happen?

config/initializers/devise.rb
```diff
@@ -309,4 +309,17 @@
...
   # changed. Defaults to true, so a user is signed in automatically after changing a password.
   # config.sign_in_after_change_password = true
+
+  # Errors from authentication may crash because no Devise mapping is defined
+  # for the callback path. To work around this we manually set the failure
+  # callback to Users::OmniauthCallbacksController.failure().
+  #
+  # See https://github.com/plataformatec/devise/issues/2004#issuecomment-7466322
+  # for more information.
+  OmniAuth.config.on_failure = proc do |env|
```
stevebaxter: This really might not be the right place to put this, however this does seem to work!

PR #334, opened 2019-12-10
Issue #324 share email address

### LINKED ISSUE #324: voter exchange of emails at swap confirm time
Possible solution to #281 @aspiers @tdg 

The essentials:

- swap proposer gives consent (or not) to share their email address at time of proposing the swap
- swap recipient gives consent (or not) to share their email address at time of confirming the swap
- when swap is confirmed, any emails consented to will be sent from SMV standard address containing email address, twitter profile and fb link with a big caveat.

Since we're only using one authentication method (twitter, email, FB) there should be no ambiguity about which email address we're using, only need a checkbox on the form.

The swap form (propose or confirm) will need to explain the need for this consent for anyone authenticating with FB.

baob:
Closes #324 . Addresses part or all of #102 (tbd)

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -199,8 +199,8 @@ class User < ApplicationRecord
 
   def confirm_swap
     incoming_swap.update(confirmed: true)
-    UserMailer.swap_confirmed(self, swapped_with).deliver_now
-    UserMailer.swap_confirmed(swapped_with, self).deliver_now
+    UserMailer.swap_confirmed(self, swapped_with, incoming_swap.consent_share_email_chooser).deliver_now
+    UserMailer.swap_confirmed(swapped_with, self, incoming_swap.consent_share_email_chosen).deliver_now
   end
 
   def clear_swap
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -150,7 +150,7 @@ class User < ApplicationRecord
     mobile_phone&.verified
   end
 
-  def swap_with_user_id(user_id)
+  def swap_with_user_id(user_id, consent_share_email)
     other_user = User.find(user_id)
     return unless can_swap_with?(other_user)
 
@@ -159,7 +159,11 @@ class User < ApplicationRecord
 
     UserMailer.confirm_swap(other_user, self).deliver_now
 
-    create_outgoing_swap chosen_user: other_user, confirmed: false
+    create_outgoing_swap(
+      chosen_user: other_user,
+      confirmed: false,
+      consent_share_email_chooser: (consent_share_email || false)
+    )
     save
   end
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -207,6 +207,13 @@ class User < ApplicationRecord
     UserMailer.swap_confirmed(swapped_with, self, incoming_swap.consent_share_email_chosen).deliver_now
   end
 
+  def swap_email_consent?
+    # Check if our swapper has given permission for us to see their email
+    return outgoing_swap.consent_share_email_chosen if outgoing_swap
+    return incoming_swap.consent_share_email_chooser if incoming_swap
+    return false
+  end
+
   def clear_swap
     if incoming_swap
       incoming_swap.destroy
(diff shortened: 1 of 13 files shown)


app/views/user/swaps/new.html.haml
```diff
@@ -5,8 +5,20 @@
...
     %p.text-center
       Are you sure you would like to swap your vote with #{@swap_with.redacted_name}?
+
     %p.text-center
-      = link_to "Yes", user_swap_path(user_id: @swap_with.id), method: :post, class: "btn btn-primary"
+      = form_tag user_swap_path, class: "form-inline" do
+
+        = hidden_field_tag 'user_id', @swap_with.id
+        %p.text-center.small
+          Please check the box if you are willing to share your
```
stevebaxter: Can we use a BS form here, put the checkbox before the label and link the label to the checkbox (so the click area is a bit bigger)? I think I might miss this if I wasn't careful.

stevebaxter: I can look at this if you want?

baob: Please  :)

baob:
I wonder if we should default the checkbox to "consent is given"  .... or is that a bit sneaky ?

stevebaxter:
We can't unfortunately, not permitted under GDPR

baob:
> We can't unfortunately, not permitted under GDPR

Thought that might be the case

app/models/user.rb
```diff
@@ -207,6 +207,13 @@ def confirm_swap(swap_params)
     UserMailer.swap_confirmed(swapped_with, self, incoming_swap.consent_share_email_chosen).deliver_now
   end
 
+  def swap_email_consent?
+    # Check if our swapper has given permission for us to see their email
+    return outgoing_swap.consent_share_email_chosen if outgoing_swap
+    return incoming_swap.consent_share_email_chooser if incoming_swap
+    return false
+  end
```
baob: 👍 yesssss  thank you

app/models/user.rb
```diff
@@ -159,7 +159,11 @@ def swap_with_user_id(user_id)
 
     UserMailer.confirm_swap(other_user, self).deliver_now
 
-    create_outgoing_swap chosen_user: other_user, confirmed: false
+    create_outgoing_swap(
+      chosen_user: other_user,
+      confirmed: false,
+      consent_share_email_chooser: (consent_share_email || false)
```
aspiers: No need for the parens here ;-)

app/views/user/swaps/new.html.haml
```diff
@@ -5,8 +5,21 @@
...
+
+        .form-check.row.mx-auto
+          %input.form-check-input{ type: "checkbox", id: "consent-check",
+                                  name: "consent_share_email_chooser" }
+          %label.form-check-label{ for: "consent-check" }
+            Share my email address with #{@swap_with.redacted_name}
+            when the swap is confirmed
+
+        .form-group.row.mx-auto
+          = button_tag "Swap with #{@swap_with.redacted_name}", class: 'btn btn-primary', type: "submit"
```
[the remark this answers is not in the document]
aspiers: HAML lets you wrap lines at commas BTW ;-)

app/views/user_mailer/swap_confirmed.html.haml
```diff
@@ -1,31 +1,34 @@
...
 %p
-	We'll send some reminders to #{@swap_with.name} to remind them to vote for your chosen party on December 12th, and we'll ask them for confirmation once they have.
+  We'll send some reminders to #{@swap_with.name} to remind them to vote for your chosen party on December 12th, and we'll ask them for confirmation once they have.
 
-	- if @swap_with.profile_url.present?
-		In the meantime, why not
-		%a{href: @swap_with.profile_url} reach out to #{@swap_with.name} directly?
+  - if @swap_with.profile_url.present?
+    In the meantime, why not
+    %a{href: @swap_with.profile_url} reach out to #{@swap_with.name} directly?
```
aspiers: No, you can't assume we have their social profile - see #192.  I see that this is actually a regression I missed :-(  I think it was introduced as part of the monster email login PR.

PR #349, opened 2019-12-11
Automatically give consent for email login

aspiers:
This is problematic because it currently asks for consent even
when email login is used.  I think this is a regression.

### DIFF
--- a/app/helpers/users_helper.rb
+++ b/app/helpers/users_helper.rb
@@ -6,7 +6,7 @@ module UsersHelper
   def contact_methods(user)
     methods = social_contact_methods(user)
 
-    if user.email.present? && user.email_consent?
+    if user.email.present? && (user.email_consent? || user.email_login?)
       methods << "by email at " + link_to(user.email, user.email_url)
     end
 
--- a/app/views/users/show/_swap_confirmed.html.haml
+++ b/app/views/users/show/_swap_confirmed.html.haml
@@ -30,6 +30,8 @@
     != methods.join " or "
 
     to get to know your swap partner.
+
   - else
-    Unfortunately #{@user.swapped_with.name} has not shared their email address or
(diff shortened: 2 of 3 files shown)


aspiers:
@baob @tdg Was this deliberate or a regression?  We didn't ask users logging in via email for consent to share their email in the past.

tdg:
Sorry - I don’t know or understand what is at play here, or remember the history or am able to find it. 

Commenting so you know I’ve seen it. 

I agree with what you said elsewhere about simplifying consent options (not needing 6 distinct models)

baob:
Not sure I understand your motivation for this change @aspiers . Admittedly I don't the know the history of our various policies ... but are we _really_ saying that if a user authenticated with email, then that email address is automatically shareable ?  I would be surprised to find any site that operated that policy ... more typically I see "this is private, we will never share your email".

But it's not my decision to make obviously ... happy for you and @stevebaxter and @tdg to work it out.

aspiers:
To be clear, this change could be total bull, which is why it's only a draft :-) Wanted to get your feedback.

aspiers:
[**@baob**](https://github.com/baob) commented on [December 11, 2019 6:24 AM](https://github.com/swapmyvote/swapmyvote/pull/349#issuecomment-564348063 "2019-12-11T02:24:54Z - Replied by Github Reply Comments"):
> Not sure I understand your motivation for this change @aspiers . Admittedly I don't the know the history of our various policies ... but are we _really_ saying that if a user authenticated with email, then that email address is automatically shareable ?  I would be surprised to find any site that operated that policy ... more typically I see "this is private, we will never share your email".

Well we reworded copy explicitly to cover that email login case, but I'm inclined to agree this is a bit contentious and hopefully not needed so much now that we have nice consent options.  Closing for now.

PR #354, opened 2019-12-11
redirect straight to swap page when there are swaps available

### LINKED ISSUE #351: Disable the share screen in the flow immediately once logged in
I think we’re giving a lot of people the wrong impression with 
“We’re looking for someone for you to swap with” and losing them out of the funnel

Is this an easy task? If so, let's disable it for a period and see if we get more proposed swaps going on

alanjc7:
Closes #351 

Currently we go to the share via social page with a message that could make it look like there are no available swaps. 
We want to try dropping this page and going straight to the swap page when there are available swaps. 

I believe the change is as simple as this but may have missed something - can't test it locally as my db is not up to date. Would be great if someone could pull this branch as run through the flow.

### DIFF
--- a/app/controllers/user/constituencies_controller.rb
+++ b/app/controllers/user/constituencies_controller.rb
@@ -10,7 +10,7 @@ class User::ConstituenciesController < ApplicationController
 
   def update
     @user.update(user_params)
-    redirect_to user_share_path
+    redirect_to user_swap_path
   end
 
   private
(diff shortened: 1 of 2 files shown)


egglestn:
Pulled down the branch and all seems fine

alanjc7:
Have also managed to get this tested locally now. Flow works as expected.
You will still land on the share screen if there are no potential swaps

PR #355, opened 2019-12-11
Issue #337 retrospective email consent

### LINKED ISSUE #337: allow swappers to retrospectively consent to sharing email
Follows on from #324. We need a mechanism for users to go back to the swap and consent to their email being shared. 

Minimum:

* page shown on confirmed swap needs to show opportunity to give consent if not yet given
* page shown on confirmed needs to notify the other party on change, notification can include the email address

* page shown on pending swap needs to show opportunity to give consent if not yet given
* page shown on pending swap should NOT send notification. That happens when confirmed.

baob:
Closes #337

Allows swappers with either an unconfirmed or confirmed swap to give consent to email sharing.
Email notification will be sent if the change is made after confirmation, otherwise the confirmation email covers it.

### DIFF
--- a/spec/controllers/user/swaps_controller_spec.rb
+++ b/spec/controllers/user/swaps_controller_spec.rb
@@ -70,48 +70,76 @@ RSpec.describe User::SwapsController, type: :controller do
         end
       end
 
-      context "when all the ducks are lined up but swap is not yet confirmed" do
-        let(:swap)  { Swap.create(chosen_user_id: swap_user.id) }
+      describe "PUT #update" do
+        context "when all the ducks are lined up" do
+          before do
+            new_user.save! # seemed to be necessary to get all the foreign keys right ... maybe try deleting later
+            new_user.incoming_swap = swap
+            new_user.save! # seemed to be necessary to get all the foreign keys right ... maybe try deleting later
 
-        before do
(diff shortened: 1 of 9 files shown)


spec/controllers/user/swaps_controller_spec.rb
```diff
@@ -71,26 +71,183 @@
...
-          expect(swap_user.swap.confirmed).to be true
+        context "when the user has an INcoming swap" do
+          before do
+            new_user.save! # seemed to be necessary to get all the foreign keys right ... maybe try deleting later
+            new_user.incoming_swap = swap
+            new_user.save! # seemed to be necessary to get all the foreign keys right ... maybe try deleting later
+
+            swap_user.save! # seemed to be necessary to get all the foreign keys right ... maybe try deleting later
+            swap_user.outgoing_swap = swap
+            swap_user.save! # seemed to be necessary to get all the foreign keys right ... maybe try deleting later
```
egglestn: You should be able to get rid of these, if you use this instead: allow(swap).to receive(:choosing_user).and_return(new_user), from memory they probably need to be inside the it / scenario, not at context scope

baob: Gah ... I totally forgot about those comments ... I will see if I can do better.

Issue #353, opened 2019-12-11
Improve message where account already exists with different login method


stevebaxter:
If someone has logged in with FB and they then try and log in with Twitter or using email, they get a message:

> Email already taken

We can improve this by showing a more informative message, e.g.:

> A user with an email address of xxx@yyy.com has already signed up using Facebook

PR #358, opened 2019-12-11
Improve email address validation behaviour

stevebaxter:
This adds a custom validation message when trying to create an account if it already exists. We tell the user how the original account was created.

This should also fix a problem some users may have when trying to log in with FB or Twitter if there are duplicate email addresses in the system.

fixes: #353 

Looks a bit like this:

Screenshot 2019-12-11 15 56 50

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -27,7 +27,9 @@ class User < ApplicationRecord
   after_save :send_welcome_email, if: :needs_welcome_email?
   before_destroy :clear_swap
 
+  # Additional validation for emails, :validatable has already added basic validation
   validate :email_uniqueness, on: :create
+
   validates :name, presence: true
 
   def omniauth_tokens(auth)
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -27,7 +27,7 @@ class User < ApplicationRecord
   after_save :send_welcome_email, if: :needs_welcome_email?
   before_destroy :clear_swap
 
-  validates :email, uniqueness: { case_sensitive: false }, allow_nil: true
+  validate :email_uniqueness
   validates :name, presence: true
 
   def omniauth_tokens(auth)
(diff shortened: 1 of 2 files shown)


app/models/user.rb
```diff
@@ -27,7 +27,9 @@ class User < ApplicationRecord
   after_save :send_welcome_email, if: :needs_welcome_email?
   before_destroy :clear_swap
 
-  validates :email, uniqueness: { case_sensitive: false }, allow_nil: true
+  # Additional validation for emails, :validatable has already added basic validation
+  validate :email_uniqueness, on: :create
```
stevebaxter: We only validate on create, which should prevent some of the problems we're seeing. `:validatable` only validates on change:

```
        base.class_eval do
          validates_presence_of   :email, if: :email_required?
          if Devise.activerecord51?
            validates_uniqueness_of :email, allow_blank: true, case_sensitive: true, if: :will_save_change_to_email?
            validates_format_of     :email, with: email_regexp, allow_blank: true, if: :will_save_change_to_email?
          else
            validates_uniqueness_of :email, allow_blank: true, if: :email_changed?
            validates_format_of     :email, with: email_regexp, allow_blank: true, if: :email_changed?
          end

          validates_presence_of     :password, if: :password_required?
          validates_confirmation_of :password, if: :password_required?
          validates_length_of       :password, within: password_length, allow_blank: true
        end
      end
```

app/models/user.rb
```diff
@@ -321,4 +323,29 @@ def test_user_suffix
   def password_required?
     false
   end
+
+  def email_uniqueness
```
stevebaxter: Custom validator. There may be a terser way to do this...

app/models/user.rb
```diff
@@ -321,4 +323,29 @@ def test_user_suffix
...
+    end
+
+    return unless existing_user
+
+    # Delete the default message that gets added
+    errors.delete(:email)
+
+    # Add our custom one
+    if existing_user.email_login?
+      errors.add(:base, "A user with this email address already exists")
```
stevebaxter: @tdg - are these messages ok?

app/models/user.rb
```diff
@@ -321,4 +323,29 @@ def test_user_suffix
...
     false
   end
+
+  def email_uniqueness
+    # Ignore nil email addresses
+    return unless email
+
+    if id
+      # Ignore self in the uniqueness check
+      existing_user = User.where("lower(#{:email}) = ? and id <> ?", email.downcase, id).first
```
aspiers:
```suggestion
      existing_user = User.find_by("lower(#{:email}) = ? and id <> ?", email.downcase, id)
```

app/models/user.rb
aspiers: Fine by me

stevebaxter:
Merged from master and resolved conflicts

PR #367, opened 2019-12-12
collapse SWAPS_CLOSED and VOTING_OPEN into SWAPMYVOTE_MODE

aspiers:
Keeping track of the logic combining SWAPS_CLOSED, VOTING_OPEN, and the sesame mechanism was getting too confusing.  So replace the env variables with a single SWAPMYVOTE_MODE variable which expresses which part of the election cycle we're in, and allow sesame to override that.

Then tweak the copy and flows to support the new `open-and-voting` mode.


baob:
Only one bug discovered on staging, and that only applies to voting day, and it was the result of an earlier change of mine, not this branch.  Merging.

PR #446, opened 2021-06-12
Simplify BE2021 fixtures to only use class methods

aspiers:


aspiers:
@baob Let me know if this makes sense or if I missed some reason why we shouldn't do this - thanks!

Guardfile
```diff
@@ -81,6 +81,7 @@ group :all_plugins, halt_on_fail: true do
     watch(rspec.spec_support) { rspec.spec_dir }
     watch(rspec.spec_files)
     watch(%{db/fixtures})
+    watch(%r{^db/fixtures/.*}) { "#{rspec.spec_dir}/db/fixtures" }
```
baob, line 84: Good :)

baob (approved):
Like it. I realised there was a lot of dead weight in what I'd done when I finished, but was in a hurry to get the code in.

PR #455, opened 2021-06-13
Swap prevention with same party

### LINKED ISSUE #70: Prevent swaps with same party

baob:
Supersedes PR #279 - It's the same branch as #279, but rebased on current master.  Should resolve #70

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -29,6 +29,7 @@ class UsersController < ApplicationController
 
   def update
     @user.update(user_params) if params[:user]
+
     if !phone_param.blank? && @user.mobile_number != phone_param
       begin
         @user.mobile_number = phone_param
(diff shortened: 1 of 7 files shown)


baob:
Comments from @aspiers on the original PR #279 

> Many thanks - this looks excellent! although I have to apologise for not making the original issue description comprehensive enough: currently, assigning willing/preferred party to the same value is actually used as a mechanism for indicating that the user doesn't want to swap:
> 
> https://www.swapmyvote.uk/faq#deactivate
> 
> This is clearly not ideal UX design but nevertheless we can't prevent it until we have an alternative mechanism for disabling swaps (and I think deleting the account is not really good enough because that requires people to reenter info and spams them with another welcome email when they rejoin).
> 
> So can I reluctantly and cheekily ask you to add a new `hidden` field to the `User` model which would allow users to temporarily pause swapping on their account, plus a toggle switch to control it from `/users/edit`? When hiding is toggled on, the code needs to call `#clear_swap` to cancel any existing swap with the account. And of course the FAQ will need to be updated accordingly.

baob:
@aspiers Really need your input on the UI. for 'swaps_paused' (as suggested by you above) I could implement a check box for 'swaps paused' but that doesn't sound very slick .. maybe it will do for the moment though.

aspiers:
@baob I found a couple of bugs with the Javascript which I fixed.  Also rebased, and squashed some related changes together.

PR #464, opened 2021-06-15
do not allow swaps in same constituency

### LINKED ISSUE #452: Stop showing candidates from the user's constituency
This seems like a regression, or maybe we just never noticed before because there were so many matches from all around the country?  @baob Any thoughts?

baob:
Fixes #452


aspiers (approved):
Amazing!

aspiers (approved):

Issue #303, opened 2019-12-07
Get Involved Page


tdg:
New static page which explains different ways (technical and non-technical) people can help the project

Github
Donate
Share
Write blogs
Upload videos with swap partners
Etc

tdg:
Expands upon #185

aspiers:
@joshr @tdg Is M8 correct for this, or is that too late?

PR #469, opened 2021-06-16
WIP Add get-involved page (#303)

aspiers:
Remnant from bad rebase?

Came from Tom's Redacting-political-affilliations-from-mails branch, the rest of which is now merged.

When finished, will address #303.


aspiers:
This is old but do we need to resurrect?

PR #486, opened 2021-06-18
Modal fix

### LINKED ISSUE #484: Login button in closed-wind-down is broken
`onclick="$('#js-login-modal').modal(); return false"` results in:

```
(index):68 Uncaught TypeError: $(...).modal is not a function
    at HTMLAnchorElement.onclick ((index):68)
```

This is super weird because it works fine in other modes, and `.modal()` is defined by bootstrap.

JFernando122:
Fixing link so the log-in modal is properly displayed (Fixes #484 )

### DIFF
--- a/app/views/layouts/_login.html.haml
+++ b/app/views/layouts/_login.html.haml
@@ -1,6 +1,5 @@
 %span.user
-  = link_to "Already been here? Log in", "#", class: "small subdued logout",
-            onclick: "$('#js-login-modal').modal(); return false"
+  = link_to "Already been here? Log in", "#js-login-modal", class: "small subdued logout", data: {toggle: "modal"}
 .modal#js-login-modal{ tabindex: "-1", role: "dialog",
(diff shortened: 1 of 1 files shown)


aspiers (approved):
Fantastic, thanks!  So why was this working in other modes but not for `closed-and-voting`?

JFernando122:
Not exactly sure as of why would work in some places and in others don't, what i found in [Stackoverflow](https://stackoverflow.com/questions/27064176/typeerror-modal-is-not-a-function-with-bootstrap-modal) was that when there are multiple jquery instances or bootstrap was called before jquery it won't work.

aspiers:
Strange, well anyway the fix is hugely appreciated even though I don't really understand it! :laughing:

aspiers:
BTW how did you find this project? :-)

JFernando122:
By searching for the "help wanted" label :)

aspiers:
Amazing, it actually works! Thanks so much again 😁

PR #487, opened 2021-06-20
refactor parties

### LINKED ISSUE #458: Refactor fragile code used to uniquely identify parties, perhaps use Electoral commission ID
Presently this mostly relies on taking the party name and reducing it to it's 'canonical' value and matching on that. This is a bit fragile given you could  find in various data sources the conservative party identified by 
* Conservative
* Conservatives
* Con
* Conservative and Unionist party

The various sources to consider are 
* Import of Electoral calculus predictions (db/fixtures/electoral_calculus_constituencies_tsv)
* Seeding the party table with the right colours (db/seeds_for_general_election.rb)
* Any custom import, like the yaml file for the 2021 By-elections
* Tactical recommendations imports. each tactical recommendation site uses subtly different shortcodes. (app/models/recommendation.rb) 

I think we need to standardise on one of:
* Electoral commission ID. Which has the complication that a single electoral unit (Labour and co-op party) corresponds to two different EC parties, with exactly the same party name, which are in permanent electoral alliance.
* Democracy club id which appears to address the Labour/Coop problem but presently comes in two flavours deprecated, and volatile. 

/cc @aspiers @tdg

baob:
should fix #458


aspiers:
@baob Was this closed because you are intending to submit a different fix?

baob:
Essentially, yes. When it's less of an experimental toolbox and more finished I'll do another PR

PR #493, opened 2021-07-03
Issue 491 duplicate and joint parties - Electoral Commission data

### LINKED ISSUE #491: Decide how to handle duplicate EC party registrations
The Electoral Commission has two registers of parties, one for NI, and one for GB.  Some parties are apparently registered twice, by looking at the EC "Entity Name" (though they differ in registration code). These include:

 "Animal Welfare Party",
 "Common Good",
 "Conservative and Unionist Party",
 "Green Party",
 "Independent Network",
 "Libertarian Party",
 "National Front",
 "Real Democracy Party",
 "Restoration Party",
 "UK Independence Party (UKIP)",
 "UK Voice",
 "Veterans and People’s Party"
 
 Other parties are variously registered to either GB or NI
 
 Question: How do we deal with this ?
 
 We can't leave the list in this state or the dropdown will offer people 2 conservative parties.
 
 Possible solutions
 
 1. Assume the two parties are one and the same and only list one "Conservative and Unionist Party". This seems to me to be the most sensible
 
 2. List  "Conservative and Unionist Party (GB)" and "Conservative and Unionist Party (NI)" probably not an accurate reflection of reality.
 
 3. (Not really an option) Never show a list of every party, in which case the problem doesn't arise. But which means not allowing users to pick from a list of 'all parties' and not having an API (because there would be ambiguity over which party was meant by  "Conservative and Unionist Party",.
 
 TO ME IT LOOKS REALLY OBVIOUS (1) IS THE SOLUTION. But I wanted to check my thinking before going forward. 
 
 Relates to #458 
 
 /cc @tdg @aspiers

baob:
Fixes #491 .

This is a wrapper and utility for the Electoral Commission data, no data is held in tables, a hash of unique parties is constructed from a CSV downloaded from the EC.

Parties with two registrations (GB and NI) are merged into one, and so are parties with a joint description (notably Labour and Co-op). No other assumptions are made about the EC Data. 

It is anticipated that entries in the hash will be used to create entries in our party table.

### DIFF
--- a/db/fixtures/electoral_commission_parties.rb
+++ b/db/fixtures/electoral_commission_parties.rb
@@ -20,7 +20,7 @@ require "csv"
 
 module Db
   module Fixtures
-    class Db::Fixtures::ElectoralCommissionParties
+    class ElectoralCommissionParties
       include Enumerable
 
       # wikipedia page at https://en.wikipedia.org/wiki/List_of_political_parties_in_the_United_Kingdom#cite_note-1
@@ -42,9 +42,8 @@ module Db
       # The EC has a concept of unique entities, so one parliamentary entity has a unique name,
       # even though it may have more than one registration (GB and NI) for that single name.
       # This returns a hash of those entities, with all their registered descriptions, and all registrations.
-      # Also, joint descriptions are collected, so that parties in permanent alliance
-      # (e.g. Labour and Co-op) can be tracked.
-      # However, this goes a bit beyond the EC modelling, and so Labour and Co-op for instance are returned as distinct
+      # Also, joint descriptions are collected, so that parties in permanent alliance can be tracked.
+      # Labour and Co-op for instance are returned as distinct
       # entities, which can be found to work together by matching their joint description.
       def unique_entities
         return @unique_entities if defined?(@unique_entities)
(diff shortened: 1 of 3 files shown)


baob:
This gives a flavour of the data assembled by this code:

image

db/fixtures/electoral_commission_parties.rb
```diff
@@ -0,0 +1,136 @@
...
+
+      # The EC has a concept of unique entities, so one parliamentary entity has a unique name,
+      # even though it may have more than one registration (GB and NI) for that single name.
+      # This returns a hash of those entities, with all their registered descriptions, and all registrations.
+      # Also, joint descriptions are collected, so that parties in permanent alliance can be tracked.
+      # Labour and Co-op for instance are returned as distinct
+      # entities, which can be found to work together by matching their joint description.
+      def unique_entities
+        return @unique_entities if defined?(@unique_entities)
+        @unique_entities = each_with_object({}) do  |p, result|
```
aspiers, line 50: I'd suggest using a more descriptive variable name than `p`, especially given that `p` is [already defined by `Kernel`](https://rubydoc.info/stdlib/core/Kernel#p-instance_method).

db/fixtures/electoral_commission_parties.rb
```diff
@@ -0,0 +1,136 @@
...
+        end
+      end
+
+      # This is unique entities, but with parties with a permanent electoral alliance
+      # (joint descriptions, e.g. Labour and Co-op) merged into one.
+      # The name given to the merged/joint party is the joint description from the EC data.
+      # Parties which do not represent merges keep the regulated entity name
+      def unique_entities_joint_merged
+        return @unique_entities_joint_merged if defined?(@unique_entities_joint_merged)
+        @unique_entities_joint_merged = unique_entities.values.each_with_object({}) do  |p, result|
```
aspiers, line 73: Same here

aspiers (dismissed):
Looks awesome to me!  Just one tiny nitpick if you can be bothered to address it.

aspiers (approved):

PR #542, opened 2022-06-01
2022 copy updates

joshr:
*NOT tested locally


baob (approved):
There are some style warnings from 'hound' but that shouldn't stop us merging, I think.

PR #543, opened 2022-06-02
issue 347 proof read emails

### LINKED ISSUE #347: Admin interface for sending test emails to given address for proof-reading
For quicker email proof-reading, we need a new admin interface for sending a test version of a given email template (with dummy values) to a given address (defaults to address of current user but can be changed).  Should also offer the option to send test versions of *all* templates in one go.

Would be especially useful given all the emails we need to send out before GE.

baob:
closes #347 and #483 

(Actually it doesn't meet all the requirements of #347 but it will get us going)

Really, really basic email copy proof-reading, plus I stole the admin page commit from #483 

Instructions for use:

*   go to https://swapmyvotedev.herokuapp.com/
*   log in with the email address you want to test
*   go to https://swapmyvotedev.herokuapp.com/admin/
*   supply the secret password to the admin section (user swapmyvote)
*   select email proofreading from the menu
*   receive about 11 test emails

Note, in your local environment, these emails will not be sent with the
mailer, instead they will go to file tmp/mails/\<your logged in email address\>

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -6,6 +6,9 @@ class AdminController < ApplicationController
   http_basic_authenticate_with name: "swapmyvote",
                                password: ENV["ADMIN_PASSWORD"] || "secret"
 
+  def index
+  end
+
   def stats
     @user_count = User.count
     @swap_count = Swap.count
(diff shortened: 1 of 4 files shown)


aspiers (approved):
Awesome!  Next step to allow selection of which email to send?  (But keeping the "send all" option would be great.)

PR #545, opened 2022-06-03
Rubocop hound update

baob:
Hound gets a bit naggy and is inconsistent with rubocop. 
Hound config also points to a rubocop version we're not using. 
Hound complains about quite reasonable haml lines that are longer than 80 chars

Lets see if we can improve things.

### DIFF
--- a/.haml.yml
+++ b/.haml.yml
@@ -1,0 +1,70 @@
+# Whether to ignore frontmatter at the beginning of HAML documents for
+# frameworks such as Jekyll/Middleman
+skip_frontmatter: false
+
+linters:
+  AltText:
+    enabled: false
+
+  ClassAttributeWithStaticValue:
+    enabled: true
+
+  ClassesBeforeIds:
+    enabled: true
+
+  ConsecutiveComments:
+    enabled: true
+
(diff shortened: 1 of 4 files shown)


baob:
Hound gets a bit naggy and is inconsistent with rubocop. 
Hound config also points to a rubocop version we're not using. 
Hound complains about quite reasonable haml lines that are longer than 80 chars

Lets see if we can improve things.

.haml.yml
```diff
@@ -33,7 +33,7 @@ linters:
 
   LineLength:
     enabled: true
-    max: 80
+    max: 200
```
aspiers, line 36: This is too long for my tastes, but not going to slow things down based on that :-)

.hound.yml
```diff
@@ -8,3 +8,6 @@ scss:
 
 jshint:
   config_file: .jshintrc
+
+haml:
+  config_file: .haml.yml
```
aspiers, line 13: Wouldn't `.hound.haml.yml` make more sense?  But can be fixed in a follow-up if necessary.

aspiers (approved):

PR #560, opened 2022-06-05
action network form embed partial

### LINKED ISSUE #558: Update Mailchimp to Action Network
There is currently (on the live, dormant homepage) email newsletter signup for Mailchimp.
This should be changed to Action Network who now provide the service. 
Privacy and other text should be reviewed and updated where needed.

joshr:
not being included anywhere yet, but intended to replace everywhere _mailchimp.html is included.

still needs more css work. unlike mailchimp, you don't embed the html, instead js builds the form. so if you want to make changes you can only use css.

Closes #558.


tdg:
Addresses #558

aspiers (approved):

Issue #561, opened 2022-06-06
Copy review - including homepage, emails, policies, FAQ etc


tdg:
1. Live homepage should be as punchy as possible
2. Minimise repetition
3. Dormant homepage should reference the immediate future (eg by elections and late launch date, what people can do now to prepare)
4. Emails should reference the by election and be reviewed for impartiality if need be
5. Policies should reflect the platforms we're using if any have changed 
6. FAQ should reflect the current login options and how eg Trust works. 
7. Bring language up to date with culture eg Trust reference to eg having a Zoom call with your confirmed partner

tdg:
8. Review share copy for Facebook and Twitter

baob:
> 8. Review share copy for Facebook and Twitter

We might find we can hide Facebook/twitter refs programatically rather than edit the text every time we switch one on or off.

tdg:
>>We might find we can hide Facebook/twitter refs programatically rather than edit the text every time we switch one on or off.
I was wondering that, yes perhaps those pieces can be commented out where needed.

>>EDIT: in fact that might already be done ... check on the dev site on heroku, that has both twitter and facebook disabled.
I suspect you're talking about login, but 8. was intended to concern sharing the platform in a social media post

tdg:
I see we have #538 covering 8., too

baob:
For the share-by-email feature created in #631 , the copy comes from the same place as the taglines for facebook and twitter shares, in the function app_tagline here: https://github.com/swapmyvote/swapmyvote/blob/8a7a41bc64f05b0e3438b4c2cd43d8ff082b92ee/app/helpers/application_helper.rb#L91.

aspiers:
@tdg What's the status of this one now?

tdg:
Ongoing 

>

tdg:
**Welcome to Swap My Vote**

>If there is already someone who wants to swap votes with you, you will have been paired up on the site. Now that the site has introduced you to each other, it’s over to you - the system is based on trust so please feel free to contact each other **through social media**; discuss policies, news, and your hopes for the new Parliament.

delete **through social media**

baob:
Sounds like this very old ticket should be tagged GE2024 ?  Or was the wrong ticket updated ?

tdg:
**[User] would like to swap their vote with you!**

Just to check timings again, it's set to 12h at the moment

tdg:
**Your vote swap preference is in demand!**

[...]

Issue #244, opened 2019-12-02
Put consent up above all login option buttons


tdg:
when we next revisit (once email login is integrated)

tdg:
Related: #15 #243 #237 #216

aspiers:
Why is this needed? Is it really urgent?

aspiers:
@tdg

aspiers:
@tdg Please can you update the title and description to clarify what type of consent this is about, and how it relates to the other consent issues we have? It's getting quite confusing with all the different nuances!

tdg:
I don't see this as relevant any more, we seem to record consent effectively

aspiers:
Yeah think you're right.  Closing.

PR #571, opened 2022-06-07
record user's consent to be sent newsletters (#195)

### LINKED ISSUE #195: Establish way of recording consent for login email address to be sent news
We need to establish a method of recording consent for using the user’s login email address to contact them for news and other purposes (from Forward Democracy). This needs to be opt in.

This might take the form of a check box at the point that we take their email address confirming (copy to be decided). It needs to create a record in the backend indicating the user has opted in (and ideally a timestamp and URL).

Note: we already have tracking of consent to be emailed by swap partners, but that is a different thing (see #345).

baob:
fixes #195

records user consent to be sent Forward Democracy emails

- update db to record user consent for newsletters from forward democracy #195
- add checkbox to form (unstyled as yet)
- get the form value into the user db entry

### DIFF
--- a/app/views/devise/registrations/new.html.erb
+++ b/app/views/devise/registrations/new.html.erb
@@ -29,6 +29,11 @@
             <%= f.password_field :password_confirmation, autocomplete: "new-password", class: "form-control" %>
           </div>
 
+          <div class="form-group">
+            <%= f.label  :consent_news_email, "Would you like to opt-in to email updates from Forward Democracy ?", class: "mb-0" %>
+            <%= f.check_box :consent_news_email, class: "form-control" %>
+          </div>
+
           <div class="actions">
             <%= f.submit "Sign up", class: "btn btn-primary btn-block" %>
           </div>
--- a/app/views/devise/registrations/new.html.erb
+++ b/app/views/devise/registrations/new.html.erb
@@ -30,7 +30,7 @@
           </div>
 
           <div class="form-group">
-            <%= f.label  :consent_news_email, "Would you like to opt-in to email updates from Forward Democracy ?", class: "mb-0" %>
+            <%= f.label  :consent_news_email, "Opt in to email updates from Forward Democracy", class: "mb-0" %>
             <%= f.check_box :consent_news_email, class: "form-control" %>
           </div>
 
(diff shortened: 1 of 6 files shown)


app/views/devise/registrations/new.html.erb
```diff
@@ -29,6 +29,11 @@
             <%= f.password_field :password_confirmation, autocomplete: "new-password", class: "form-control" %>
           </div>
 
+          <div class="form-group">
+            <%= f.label  :consent_news_email, "Would you like to opt-in to email updates from Forward Democracy ?", class: "mb-0" %>
```
aspiers, line 33: Bearing in mind the principle "should be as punchy as possible" (e.g. as mentioned in #561):

```suggestion
            <%= f.label  :consent_news_email, "Opt in to email updates from Forward Democracy", class: "mb-0" %>
```

app/views/devise/registrations/new.html.erb
```diff
@@ -29,6 +29,11 @@
...
+            <%= f.check_box :consent_news_email, class: "form-control" %>
```
aspiers, line 34: Can we make this default to checked?  Otherwise most people won't check it, and then we miss the majority of the benefit of this PR.

aspiers (changes requested):

app/views/devise/registrations/new.html.erb
tdg: My understanding is that having it opt-in, ie unchecked by default, is part of the same legal principle that says it's needed. It makes communications marketing harder, but makes sense from the user perspective.

tdg: (Ie a large part of the benefit of the Issue is making the product better compliant)

joshr: my understanding too, i.e. it can't be pre-ticked.

specifically because the form itself is not a newsletter signup. if it was a newsletter signup you wouldn't even need the opt-in, because that's explicit.

as an aside, that's one of the issues with the action network form, it forces an opt-in on what is ONLY a newsletter signup. very frustrating.

tdg: also as an aside, #244 was partly to help maximise 'conversions', iirc

baob: Okay then, sounds like we shouldn't do this change.

baob:
@aspiers Can this be merged now ?

aspiers (approved):

PR #574, opened 2022-06-08
fix @SwapMyVote and URL appearing twice in twitter message

baob:
No github issue filed; problem raised in whatsapp.

image


baob:
After fix, looks like this:

image

PR #578, opened 2022-06-09
sign up form opt in field fix

joshr:
- added bootsrap classes for checkboxes
- reordered the checkbox and label


joshr:
should look like this now

Screenshot 2022-06-09 at 10 40 58

aspiers (approved):

PR #600, opened 2022-06-12
issue 594 my email consent confusion

### LINKED ISSUE #594: one of def my_email_consent? and def email_consent? has to be giving wrong answer
on app/models/user.rb ... complicated history with both me and adam messing with the same function in parallel branches.

image

baob:
fixes #594

- two methods with identical purpose and implementation: merge and give intent-revealing name


aspiers (approved):
Great detective work, thanks!

PR #602, opened 2022-06-12
only shrink width of profile-content if polls are shown

aspiers:
When we're in two-constituency by-election mode, we hide GE2019 polls.  But this left some unused space in the profile card, so fix the layout to only shrink the `div.profile-content` to 60% width if we're showing polls.

Also move the two party / hidden polls test from a controller to a helper, so that it's available everywhere in user views by default, which means we don't have to pass it around as a local.

### DIFF
--- a/app/assets/stylesheets/profile.scss
+++ b/app/assets/stylesheets/profile.scss
@@ -20,9 +20,11 @@
         padding-left: 76px;
         box-sizing: border-box;
 
-        @include largescreen {
-            width: 60%;
-            float: left;
+        &.with-polls {
+            @include largescreen {
+                width: 60%;
+                float: left;
+            }
         }
     }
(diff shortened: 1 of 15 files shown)


baob:
The refactoring you've done repeats what I've done in #598 (PRed yesterday). Might be a good idea to combine these two somehow before merge.

baob:
You may be able to cherry-pick https://github.com/swapmyvote/swapmyvote/commit/410ff3629cc1f7fd5ccb82eabed9d2047bb30093

baob (changes requested):
See my comment https://github.com/swapmyvote/swapmyvote/pull/602#issuecomment-1153145167

baob (approved):
Looks good to me

Issue #612, opened 2022-06-16
Craft OGP information for static pages


tdg:
To help copy and images stand out against each other and in search eg.

/FAQ could include 
> Is it legal? Yes!
> Are you independent? Yes!
> How do you trust your partner… more

An image could reflect one or more of these questions, along with branding

/API could include
> Campaigners: how to use Swap My Vote to campaign for tactical voting

An image could reflect how a link is constructed, along with branding

/About could include
> Swap My Vote helps make votes count.
> It is produced by Forward Democracy and a small group of developers who volunteer their time to make democracy a little bit better 

etc

tdg:
If it's as straightforward as I suspect, this could be part of #561

baob:
@tdg So you want to replace the text in these two tags

    <meta content='Make votes matter! Swap yours with someone in a constituency where both could count for more. You get to vote for who you really want, and help someone else do the same. Turn tactical voting into a win-win!' property='description'>
    <meta content='Make votes matter! Swap yours with someone in a constituency where both could count for more. You get to vote for who you really want, and help someone else do the same. Turn tactical voting into a win-win!' property='og:description'>

With text that's specific to a page ?

Is it just those two tags  ( og:description and description ) or are there more.

It looks not difficult at all if I just know what you want.

tdg:
There is also an image tag somewhere – that pulls the currently pink images across to social media etc

tdg:
I suppose more broadly it might be useful to be able to add specific meta information, not just OG information, to any page – an og tag being simply one of a few specific meta tags. 

Twitter, for example, recognises OG tags, but you can also specify Twitter-specific images and other cards..

But, to keep it simple, being able to implement OG information per page, that would be great as it is pretty universally recognised

tdg:
https://ogp.me/

baob:
Alright, so all pages presently already have an og:description. Presumably that must disappear conditionally on there being an override?

Or alternatively do you want me to take out the existing global one, and only have og:description where it's explicitly set per page.

And what about the plain description tag which is presently identical to the og:description tag. Should they stay in sync, or be independent?

tdg:
Let's have a principle of failing gracefully if specific page information is not set. So they would inherit an override or a global setting, and a new page that has started with also inherit the global, is that possible?

baob:
Ok that gives me something to get started, thanks

baob:
[...]

PR #629, opened 2022-06-19
issue 612 craft ogp info static pages

baob:
Gives the means to finish #612 . Placeholder text is the same as the
default. All that's needed to complete #612 is finalising the text.

 I will create a review app for this. 

- layout change to make override of meta og:description possible (#612)
- overrides for all the static pages
- placeholder overrides for all the static pages, with the same text as default, just needs edit when text is ready (#612)


baob:
@tdg deployed to a review app and ready to review.

aspiers (approved):

Issue #625, opened 2022-06-18
Ask user for constituency before showing them swaps and before asking their 'willing' party


baob:
## Motivation

* <strike>If we show them swaps without knowing their constituency then we could easily be showing them voters in their own constituency ... who they would actually never swap with. This becomes HIGHLY probable in a by election</strike> **EDIT:** I don't think this applies after doing #622 

* if we allow them to select their 'willing' party before we know their constituency, they might easily offer a party they actually cannot vote for. (previously noted as a separate problem: https://github.com/swapmyvote/swapmyvote/issues/539#issuecomment-1173628722 )

* <strike>if we allow them to select a voter to swap with before we know their constituency, then we have to interrupt their create swap flow to ask them their constituency (see e.g. bug scenario in #622) </strike> **EDIT:** I don't think this applies after doing #622 

* Not knowing consituency up front could potentially cause an engagement drop-off as suggested by #669.

## Possible solutions

* ask for constituency first, before anything else and even before registration (see https://github.com/swapmyvote/swapmyvote/issues/625#issuecomment-1173085610)

* for by election, ask them their constituency at the same time as preferred and willing party ... this just adds a 2 item drop-down which isn't too hard

* <strike>for general election, after registration, the user is expected to give their constituency after registration. This is very similar to how things are now, except instead of it being optional, we make it mandatory, and do not list swaps until it's done.</strike> (option removed to be consistent with comments in this issue and in https://github.com/swapmyvote/swapmyvote/issues/539#issuecomment-1173637017 )

@tdg @aspiers @joshr thoughts please.

baob:
Added 'Help wanted' tag ... we need a team decision on this.

@tdg @aspiers what do you think.

baob:
See also #636

aspiers:
I think we should almost certainly ask for constituency as the very first thing, before asking for anything else.  That way we can later show much more precise info about which parties and even candidates they are able to vote for, when asking for their "willing" choice.  It could even be a list with checkboxes if we support multiple willing parties as per #2.

aspiers:
BTW @baob the "help wanted" label is for attracting the attention of drive-by contributors (IIRC this is a standard across the whole of GitHub) so we don't want that here.  I think the label you are looking for is "needs refinement".

baob:
Description and title updated to reflect recent comments in https://github.com/swapmyvote/swapmyvote/issues/539#issuecomment-1173628722

baob:
[...]

Issue #618, opened 2022-06-17
mailgun config on heroku review apps broken


baob:
What we presently have on review is minimally useful ... but anything that requires email functionality fails. E.g. registering, setting up swaps.

aspiers:
Seems reasonably important to uncripple review apps, so moving to M8.

PR #633, opened 2022-06-20
close off the route to empty constituencies via user/constituencies/edit (#622)

### LINKED ISSUE #622: Bug: user asked to select constituency on creating swap, but allowed to leave blank
The scenario is
- user registers without specifying constituency
- user is offered swaps (both constituencies, since we don't know where they are)
- user chooses a potential swap
- app knows that constituency is missing and asks user to choose (check introduced in #289) 
- user hits save WITHOUT CHOOSING constituency and the APP ALLOWS IT.

Discovered while investigating #621

baob:
Should close #622

Please can someone other than me try to prove this works ? @tdg @aspiers ?

Register as a new user, and see if you can get to a point of seeing
swaps without setting your constituency ?

I don't think it can be done any more, as of this fix

### DIFF
--- a/app/controllers/user/constituencies_controller.rb
+++ b/app/controllers/user/constituencies_controller.rb
@@ -11,14 +11,14 @@ class User::ConstituenciesController < ApplicationController
   def update
     ons_id = user_params[:constituency_ons_id]
 
-    if (!ons_id.nil? && !ons_id.empty?)
+    if !ons_id.nil? && !ons_id.empty?
 
       @user.update(user_params)
       redirect_to user_swap_path
 
     else
 
-      flash.now[:errors] = ['You must tell us your constituency. Without it, the swaps we offer may not make sense']
+      flash.now[:errors] = ["You must tell us your constituency. Without it, the swaps we offer may not make sense"]
       edit
       render "edit"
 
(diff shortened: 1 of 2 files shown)


tdg:
image

I got this error message at https://swapmyvote-pr-633.herokuapp.com/users 

I was using one of my principal email addresses (@m……….tv) which might exist already in the database?

tdg:
Replicated error with two other email addresses, one of which was definitely unique

baob:
Ohhh ... totally forgot, anything that uses email (including signing up as a new user, which I asked you to do) fails on the review apps. There's a GitHub issue for it.

I'd better think of another plan for testing.

baob:
@tdg and @aspiers could you please test on your own, local apps if possible, by checking out the branch constituency-is-not-optional .

Testing on review apps will be hard in this case due to the problem raised in #618 .

tdg (approved):
Every time I try to save edit user without a constituency it takes me to that page, and the error message works there. It seems like the only way to leave it blank would be to ignore what we're asking and simply leave it black, so it works as increasing likelihood of user providing the constituency.

The only way I can imagine ensuring every user specifies constituency would be to do so on first sign up perhaps as part of the initial action flow

baob:
> The only way I can imagine ensuring every user specifies constituency would be to do so on first sign up perhaps as part of the initial action flow

That's pretty much what I was saying in https://github.com/swapmyvote/swapmyvote/issues/625

Issue #656, opened 2022-07-08
Automate election name/changes changes to minimise manual edits for a new election


baob:
We have about 10 different places on the site where the name of the election (currently Tiverton....) is mentioned. it would be good to put in place code that will automate the changes.

E.g.
* If we have 2 constituencies in the constituencies data, the name of the election is "Constituency X and Constituency Y" (and variations for hashtags etc)
* Otherwise the text is "General Election"

Automating the changes is relatively easy for General elections and by-elections that happen on the same day, but harder for a by-election with staggered dates (as in 2021).

We should consider exactly how (and  even if) we deal with staggered dates. Two possible approaches if we do it.

* model the date of each election separately and change the text on the different 'mode' pages accordingly (e.g. "constituency X has voted and now we're locked until the polling day in constituency Y")
* have separate ENV variables for each instance where the text "Constituency X and Constituency Y" will not do the job.

baob:
PR #659 adds two helper methods in application_helper.rb that should help here.

baob:
This seems to have been completed. See https://github.com/swapmyvote/swapmyvote/blob/master/app/helpers/application_helper.rb#L114 

PR #674

PR #674, opened 2022-08-11
Automate election name/changes changes to minimise manual edits for a new election

baob:
This closes the issue of the same name, #656

* The name of the by-election (if it is a by-election) is automated from
  the constituencies
* the environment variable ELECTION_DATE needs to be set to choose the
  election date
* the environment variable ELECTION_TYPE is provided to set the election
  type to general or by-election
* documentation has been updated accordingly

Review app now available at https://swapmyvote-pr-674.herokuapp.com/

Set up for full general election, no users at present


app.json
```diff
@@ -76,6 +76,16 @@
...
       "required": "false",
       "value": ""
+    },
+    "ELECTION_DATE": {
+      "description": "date of the election in the format YYYY-MM-DD",
+      "required": "false",
+      "value": ""
+    },
+    "ELECTION_TYPE": {
+      "description": "Type of election, general or by / g or b",
```
aspiers, line 86: I think it should just be `general` or `by-election`.  It's too easy to make a typo between `g` and `b` (the keys are right next to each other) so requiring the full word introduces some error-correction bits.  Additionally, it helps others if we avoid somewhat cryptic values / identifiers.

app/helpers/application_helper.rb
```diff
@@ -88,12 +88,138 @@ def donate_info
...
+      end
+      options = { words_connector: ", ", last_word_connector: " or ", two_words_connector: " or " }
+      "#{hashtags.to_sentence(options)} #byelection"
+    end
+  end
+
+  def election_date
+    date_from_env = ENV["ELECTION_DATE"]
+    return Date.parse(date_from_env) unless date_from_env.nil? || (date_from_env == "")
+    Date.parse("2022-6-23")
```
aspiers, line 140: Looks like some debugging which got left in?

app/helpers/application_helper.rb
```diff
@@ -88,12 +88,138 @@ def donate_info
...
+    when 1..2 then :winter
+    when 3..5 then :spring
+    when 6..8 then :summer
+    when 9..11 then :autumn
+    when 12 then :winter
+    end
+  end
+
+  def election_date_md
+    # "June 23rd"
```
aspiers, line 188: This is the American version, so do we ever want this?  Could just change them all to the `_dm` version.

app/helpers/application_helper.rb
```diff
@@ -88,12 +88,138 @@ def donate_info
...
+      "another constituency"
+    else
+      "the other constituency"
+    end
+  end
+
+  def election_type_override
+    election_type = ENV["ELECTION_TYPE"]
+    return nil unless !election_type.nil? && election_type.size.positive?
+    return :general if election_type[0].downcase == "g"
```
aspiers, line 214: As mentioned in a previous comment I would prefer to only allow `general` and `by-election` (or `by election` if the hyphen is incorrect), and raise an error if any other value is given.

app/views/user_mailer/welcome_email.html.haml
```diff
@@ -9,8 +9,12 @@
...
 %p
 
-  The 2022 summer by-elections for Wakefield, and Tiverton & Honiton
-  are on <em>Thursday, 23rd June</em>.
+  - if general_election?
+    The #{election_date_season_type} is
+  - else
+    The #{election_date_season_type} for #{by_election_constituencies_as_sentence} are
+
+  on <em>Thursday, #{election_date_dm}</em>.
```
aspiers, line 17: It might not be a Thursday ;-)

doc/new_election/by_election_data_coding_tasks.md
```diff
@@ -4,23 +4,24 @@ In general [db/fixtures/README](../../db/fixtures/README.md) can be a guide. If
...
+    db/fixtures/be2022_yaml.rb
     spec/db/fixtures/be2022/candidate_spec.rb
     spec/db/fixtures/be2022/party_spec.rb
-    db/fixtures/be2022.yml
-    db/fixtures/be2022_yaml.rb
     spec/db/fixtures/be2022_yaml_spec.rb
 
+### Switch to by election mode if need be
+
+See [db/seeds.rb](../../db/seeds.rb) and optionally edit so that the default election type is by election. This can be overridden using an environment variable (ELECTION_TYPE=by), however, making the edit means other developers don't need to remember to use the envirobment variables.
```
aspiers, line 23:
```suggestion
See [`db/seeds.rb`](../../db/seeds.rb) and optionally edit so that the default election type is by election. This can be overridden using an environment variable (`ELECTION_TYPE=by`), however, making the edit means other developers don't need to remember to use the environment variables.
```

doc/new_election/general_election_data_coding_tasks.md
```diff
@@ -6,11 +6,11 @@ In general [db/fixtures/README](../../db/fixtures/README.md) can be a guide. If
 
 ### Switch to general election mode if need be
 
-See [db/seeds.rb](../../db/seeds.rb) and [db/seeds_for_general_election.rb](../../db/seeds_for_general_election.rb) and edit.
+See [db/seeds.rb](../../db/seeds.rb) and optionally edit so that the default election type is general election. This can be overridden using an environment variable (ELECTION_TYPE=general), however, making the edit means other developers don't need to remember to use the envirobment variables.
```
aspiers, line 9:
```suggestion
See [`db/seeds.rb`](../../db/seeds.rb) and optionally edit so that the default election type is general election. This can be overridden using an environment variable (`ELECTION_TYPE=general`), however, making the edit means other developers don't need to remember to use the environment variables.
```

aspiers (approved):
Awesome! Couldn't find anything of any significance to suggest - will merge.

aspiers:
Actually the "Thursday" mention does need fixing.  The others are nice-to-have but we can certainly live without.

app/helpers/application_helper.rb
baob: No, that was meant to be a default, for any environment (e.g. heroku previews ?) where we forget to set ELECTION_DATE .

app/views/user_mailer/welcome_email.html.haml
baob: Seems pretty unlikely not to be though :). "Since 1918 a General Election has always been on a Thursday, except for 1918, 1922, 1924 and 1931. ". In the rare event that it is not, we can either automate the change or do it manually . 

https://www.theguardian.com/notesandqueries/query/0,5753,-2036,00.html

app/helpers/application_helper.rb
aspiers: Ah I see.  I would have a very mild preference for it failing to start up if the variable was missing, as this would alert us to the incorrect configuration rather than upgrading Heroku workers to something with the wrong date... but it's a tiny corner case and really doesn't matter :-)

app/views/user_mailer/welcome_email.html.haml
aspiers: Oh interesting, I didn't know that!  Is that also true for by elections too?

doc/new_election/general_election_data_coding_tasks.md
aspiers: Addressed in #784.

app/views/user_mailer/welcome_email.html.haml
aspiers: Follow-up in #686.

doc/new_election/by_election_data_coding_tasks.md
aspiers: Fixed in #784.

aspiers:
Follow-up in #686.

PR #723, opened 2023-09-30
issue 722 party names inconsistent

### LINKED ISSUE #722: Party names are inconsistent across text files and code

baob:
Use party names consistently.

I think in the case of the newly added reform party there was a risk of two parties being created, depending on which scripts were run.

- consistently use conservatives rather than a mix of plural and singular
- consistently use green rather than green party
- consistently use reform rather than reform party or reform uk

Closes #722

To see the update you will need to run

    bundle exec rake db:schema:load
    ELECTION_TYPE=g bundle exec rake db:seed

THIS WILL DESTROY YOUR CURRENT DEV DATABASE !

Not ideal maybe, but we're at a stage in the election cycle where I don't think this matters much.

### DIFF
--- a/app/models/party.rb
+++ b/app/models/party.rb
@@ -5,7 +5,7 @@ class Party < ApplicationRecord
     lab: { name: "Labour", color: "#DC241f" },
     libdem: { name: "Liberal Democrats", color: "#FFB602" },
     green: { name: "Green", color: "#6AB023" },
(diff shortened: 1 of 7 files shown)


baob:
**Changes to party drop-down**

image


**Changes to API dos**

image

aspiers (approved):

baob:
@aspiers can we merge this now ? You already approved.

Issue #754, opened 2024-05-27
integrate STT 'heart' recommendation


baob:
Latest from STT

> updated the CSV for the latest version so you either get: 

> •⁠  ⁠Where there's party advice: Party short code ("Lab", "LD", "Grn", "PC", "SNP") 
•⁠  ⁠Where we think Tories & RefUK can't win "Heart" 
•⁠  ⁠Where we're not giving advice yet it's blank.

> "None" is not being used anymore.

> There is a separate column now to differentiate no advice because we're not ready to release it vs we think it's too difficult to call.

baob:
Closed by #757

PR #757, opened 2024-05-27
issue 754 stt heart

baob:
Closes issue #754

STT tactical vote recommendation - "Heart" update

This is necessarily built on the branch for previous PR https://github.com/swapmyvote/swapmyvote/pull/762 . For ease of review, that. should be reviewed and merged first.

STT said:
>  ⁠Where we think Tories & RefUK can't win "Heart"

Main steps

- new STT csv
- update advice key (extra columns in the CSV)
- STT: only allow through Heart and party names

EDIT: Superseded image removed, see latest comment.

### DIFF
--- a/db/fixtures/tactical_vote_stt_csv.rb
+++ b/db/fixtures/tactical_vote_stt_csv.rb
@@ -17,7 +17,7 @@ class TacticalVoteCsv
   def initialize
     @file_name = FILE_NAME
     @link = "https://stopthetories.vote/"
-    @site = "StopTheTories"
+    @site = "stop-the-tories"
   end
 
   def data
--- a/db/fixtures/tactical_vote_stt_recs.rb
+++ b/db/fixtures/tactical_vote_stt_recs.rb
@@ -8,6 +8,7 @@ class TacticalVoteSttRecs
   attr_reader :advisor, :mysoc_constituencies
 
   ACCEPTABLE_NON_PARTY_ADVICE = ["Heart"]
+  ADVICE_TRANSLATION = { "Heart" => "Any" }
 
   def initialize
     @advisor = TacticalVoteCsv.new
@@ -28,10 +29,18 @@ class TacticalVoteSttRecs
       ons_id = ons_id_by_mysoc_short_code[row[:mysoc_short_code]]
       rec_key = { constituency_ons_id: ons_id, site: advisor.site }
       rec = Recommendation.find_or_initialize_by(rec_key)
-      rec.text = row[:advice]
 
-      party_short_code = rec.party_short_code_from_text
-      acceptable = party_short_code || ACCEPTABLE_NON_PARTY_ADVICE.include?(rec.text)
+      source_advice = row[:advice]
+      advice_is_not_party = ACCEPTABLE_NON_PARTY_ADVICE.include?(source_advice)
+
+      if advice_is_not_party
+        rec.text = ADVICE_TRANSLATION[source_advice] || source_advice
+        acceptable = !!rec.text
+      else
+        rec.text = source_advice
+        party_short_code = rec.party_short_code_from_text
+        acceptable = !!party_short_code
+      end
 
       if rec.text == "" || !acceptable
         # if it's not acceptable, or blank we must delete the existing entry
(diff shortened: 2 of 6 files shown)


db/fixtures/tactical_vote_stt_recs.rb
```diff
@@ -5,9 +5,10 @@
 require_relative("mysociety_constituencies_csv")
 
 class TacticalVoteSttRecs
-  ACCEPTABLE_NON_PARTY_ADVICE = ["None"]
   attr_reader :advisor, :mysoc_constituencies
 
+  ACCEPTABLE_NON_PARTY_ADVICE = ["Heart"]
```
aspiers, line 10: IIUC this doesn't make much sense to users, shouldn't we have some human-readable text which explains what it means?

aspiers (approved):

db/fixtures/tactical_vote_stt_recs.rb
baob: I think the only options available to us are:

* change the link back to STT to a page that explains what this means (which doesn't exist yet).
* make an adjustment to the translation of the text "heart" on import.

BUT, the second option, amounts to us interpreting what STT have said (unless of course, we're just keeping up with their UI). Which is questionable, given the neutral positioning of SMV.

tdg: I’m happy with doing what amounts to a precis, just not substantively changing or interpreting the meaning. We can always link out for further context, and in fact should be doing this somehow (without driving our users away of course)

baob: What STT say is

> Where we think Tories & RefUK can't win "Heart"

and that's been explained to me as "To stop the Tories, you can safely vote with your heart, because there won't be an impact"

Another way of saying that is "Any"

**PROPOSAL**: for the "heart" constituencies, we display the text "Any" (instead of party name)

aspiers: "Any" makes a ton more sense than "Heart" so yeah let's do that for now. However I'm convinced we can make it clearer with follow-up PRs to add a tooltip or similar which links out to the relevant page for an explanation.

aspiers:
This can be rebased now.

baob:
New screenshot

image

db/fixtures/tactical_vote_stt_recs.rb
```diff
@@ -28,10 +29,19 @@ def load
...
-      rec.text = row[:advice]
 
-      party_short_code = rec.party_short_code_from_text
-      acceptable = party_short_code || ACCEPTABLE_NON_PARTY_ADVICE.include?(rec.text)
+      source_advice = row[:advice]
+      advice_is_not_party = ACCEPTABLE_NON_PARTY_ADVICE.include?(source_advice)
+
+      if advice_is_not_party
+        rec.text = ADVICE_TRANSLATION[source_advice] || source_advice
+        party_short_code = nil
```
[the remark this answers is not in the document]
baob: Hound is wrong. behaviour is different without this line.

db/fixtures/tactical_vote_stt_csv.rb
```diff
@@ -17,7 +17,7 @@ class TacticalVoteCsv
   def initialize
     @file_name = FILE_NAME
     @link = "https://stopthetories.vote/"
-    @site = "StopTheTories"
+    @site = "stop-the-tories"
```
aspiers, line 20: If this is what gets displayed in the boxes, TBH this looks a bit ugly to me, but we can debate that later ;-)

aspiers (dismissed):

aspiers (approved):

db/fixtures/tactical_vote_stt_csv.rb
baob: The reason for the hyphens, I discovered, is that if you have 7-8 recommendations (which we DID have) the hyphens nicely wrap on the horizontal version of the layout. 

But yeah, I'm sure a proper web designer could do a better job.

aspiers: I believe the correct solution to that is [a zero width space](https://stackoverflow.com/a/15841796/179332).

Issue #759, opened 2024-05-29
API page references wrong field names for ONS id and name


baob:

baob:
Closed by #760

PR #760, opened 2024-05-29
correct API help page information about field names

baob:
closes issue #759

The data source for constituencies was already swapped in a previous commit, but I missed the fact we named the fields in the data source for API users.


aspiers (approved):
Nice one.  BTW I would suggest that changes like these are squashed into a single commit, otherwise you have the first commit breaking the test suite, and the second one fixing it again:

image

This masks any other potential failures which might be lurking in the first commit, but more importantly it breaks things like `git bisect`.  Ideally every single commit should be 100% passing the test suite.

Issue #766, opened 2024-06-02
BUG: After User Edit then save, user stays on edit menu, no return to view swaps


baob:
Only happening to my account on heroku dev server. Not yet reproduced on my dev machine.

Seemed to start after I let autofill populate my postcode, which also affected the email field.

baob:
Any relation to #378 ?

aspiers:
@baob Isn't this a duplicate of #166?

aspiers:
... for which I think the best solution was already suggested by @stevebaxter: https://github.com/swapmyvote/swapmyvote/issues/166#issuecomment-2203184826

aspiers:
@baob Oh, is this issue intended to be _specifically_ about the flow when a user who hasn't entered their phone number wants to offer a swap, so they get redirected to user edit but don't get redirected back to the swap flow afterwards?

That wasn't super clear from the issue title or description, I'll edit.

aspiers:
Actually I really need clarification here... should "no return to view swaps" actually be "no return to _offer_ a swap"?

baob:
> should "no return to view swaps" actually be "no return to _offer_ a swap"?

Looking at the context, yes, I'm sure that's what I encountered ... tried to make a swap, redirected to verify, but then no return to that same place afterwards.

Issue #166, opened 2019-11-27
/user/edit workflow needs refinement


aspiers:
In `/user/edit` (the user's profile editing view), normally the Save button will return back to the normal vote swapping workflow, so the button really means "Save and continue".  However if a mobile phone number is set and needs verification, the Save button will just save any changes to the database but stay on that page in order to give the user a chance to click the Verify button.  In that case the button really means "Save but don't continue".

This is not the most intuitive UI:

1. It means that in the "needs verification" scenario, it's not obvious how to return to vote swapping.
2. It means we have an identical-looking button with two different behaviours.

So it should be improved.  Maybe the ideal UI is to detect the case where the mobile number has just been set, and in that case *only*, offer two separate buttons: "Save" and "Save and continue".

stevebaxter:
Could we just send the code on "Save" if the number is unverified, and take the user to the verification page?

aspiers:
Haha DOH that's much simpler and more elegant isn't it!  Not sure why I thought a verify button was necessary ...  @egglestn I guess this approach might also help streamline your work, but I haven't had a chance to look at what you've done yet.

aspiers:
Please noone work on this without talking to @egglestn and me first.

baob:
See also #636

baob:
@stevebaxter is this on your radar ?

baob:
Need to be careful about the overlap with #625 here ... don't want to be treading on each others toes.

baob:
Relates to #928 ?

aspiers:
And possibly / probably #933 too.

aspiers:
@stevebaxter A long time ago [you came up with this great suggestion](https://github.com/swapmyvote/swapmyvote/issues/166#issuecomment-559479666):

> Could we just send the code on "Save" if the number is unverified, and take the user to the verification page?

Any chance you have time for this later?  It seems to be a common stumbling block still...

stevebaxter:
OK, how about this flow:

https://www.dropbox.com/s/k9b5t1reoi4x8um/Screen%20Recording%202024-07-02%20at%2022.48.44.mov?dl=0

stevebaxter:
The main differences are:

1. Swap the order of the postcode and constituency fields, and make it clear that the postcode is only to find the constituency (this has bugged me for ages)
[...]

PR #767, opened 2024-06-02
tell user they need to verify mobile number

baob:
Scenario: user has entered mobile phone number on /user/edit but hasn't yet verified.

Might happen
- right after registration (esp. if browser autofills email and phone number)
- when about to make the 1st swap, in which case the rules are we need a mobile phone number

Once on the /user/edit page with mobile phone filled in, the only way save will work is after verifying the mobile phone number. Alternatively they can navigate away to the home page.

This commit does not change that behaviour, it just lets the user know why save is not working with an error message.

This mitigates issue #766 and might be considered a fix.

It also directly addresses the problem raised in #166 though the solution is different.

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -1,5 +1,10 @@
+
 Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
+
+  # Use the lowest log level to ensure availability of diagnostic information
+  # when problems arise.
+  config.log_level = :debug
 
   # In the development environment your application's code is reloaded on
   # every request. This slows down response time but is perfect for development
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -51,7 +51,7 @@ class UsersController < ApplicationController
       return edit_user_path
     end
 
-    return edit_user_path if !@user.valid?
+    return edit_user_path unless @user.valid?
 
     # Otherwise, return to the user path - user will see a verification prompt if needed
     user_path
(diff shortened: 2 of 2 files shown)


config/environments/development.rb
```diff
@@ -1,6 +1,11 @@
+
 Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
 
+  # Use the lowest log level to ensure availability of diagnostic information
+  # when problems arise.
+  config.log_level = :debug
```
aspiers, line 7: @baob Was it intentional to include this hunk in this PR?

aspiers: I can imagine it being useful, just wondering if there are any downsides to having it as the default (e.g. if it's so verbose that it makes normal log output harder to follow)

baob: It was deliberate ... took me ages to discover what exactly it was I needed to do. So unless we can identify a downside, I'm voting for keeping it.

aspiers: OK. Would be helpful to understand the value of debug logging though, so that I and others can also benefit from it:

Do you mean it took you ages to figure out how to increase the log level, or how to trigger a flash message, or something else?

aspiers (approved):

config/environments/development.rb
baob: Took me ages to figure out the configuration option I needed to enable me to write debugging logging statements in code that actually showed up somewhere readable.

Issue #773, opened 2024-06-06
Annotate Swaps with parties in tactical vote recommendations


baob:
Examples:

Users preferred party is green, recommendation is green, we say ":tick: Good swap (according to \<recommendation site\>)
Users preferred party is green, recommendation is Any, we say ":tick: Good swap (according to \<recommendation site\>)
Users preferred party is green, recommendation is LibDem, we say nothing. 

#772 MUST be done first, presently tactical vote recommendations are purely text, we need to match that to parties specified by users to make the annotation

baob:
Overlaps with #667

baob:
Suggestions from chat elsewhere:
 - [ ] use full party names not abbreviations
 - [ ] show ''TV site had no advice' as appropriate
 - [ ] just have the tick and no "good or bad".
 - [ ] a fake RW campaign (or two) FOR USE ONLY IN DEV to balance out the LW ones.

baob:
The NI problem

![image](https://github.com/swapmyvote/swapmyvote/assets/55932/ebad6a23-6b68-4230-b868-ba3b7c36a812)

aspiers:
@baob  I suggest filing a separate issue for that NI issue - while it's not great it doesn't seem like a launch blocker.

PR #775, opened 2024-06-08
issue 772 tv advice party links

### LINKED ISSUE #772: Relate Tactical Vote recommendations to Parties in database
This is a pre-requisite to doing #773

baob:
Addresses and closes #772 Relate Tactical Vote recommendations to Parties in database

This unblocks #773 Annotate Swaps with parties in tactical vote recommendations

Built on the PR for #768, best to review that first.

# Summary of approach

New table created RecommendedParty where rows represent what one TV advisor said about one party in one constituency.

3 current TV imports were re-written to discover actual parties that the advice text recommended.

Where the text advice implies multiple parties, multiple records are created.

# Steps:

- better name: Party#smv_code
- add rails model to link parties to TV recommendations
- make recommended parties findable from constituency
- make recommendations findable from party
- method for parties standing in constituency
- ensure dont recommend Plaid outside Wales, etc
- total re-write of STT import
- extract method for recommended parties update
- total re-write of sprintforpr import
- total re-write of tactical.vote import

# To Verify

- check out branch
- get db up to date with either

    rake db:schema:load

  or

    rake db:migrate

- seed the database

    ELECTION_TYPE=g rake db:seed

- use rails console to explore

    \# all recommended parties for one constituency
    OnsConstituency.find_by_name("Torfaen").recommended_parties.all.map { |rp| [rp.site, rp.party.name] }

    \# Five constituencies where labour is recommended
    labour = Party.find_by_smv_code(:lab)
    RecommendedParty.joins(:constituency).where(party_id: labour.id ).limit(5).all.map { |rp| [rp.constituency.name, rp.party.name] }


db/seeds_for_general_election.rb
```diff
@@ -24,6 +24,13 @@
   puts party_attributes.slice(:name, :smv_code)
 end
 
+unique_smv_codes = Party.all.each_with_object(Set.new) { |p, set| set.add(p.smv_code) }
```
aspiers, line 27: Nevertheless, it wouldn't cost anything significant to add a db constraint for this too, would it?  I would have thought better to have defence in depth.

db/fixtures/tactical_vote_stt_recs.rb
```diff
@@ -31,31 +41,48 @@ def load
...
-        acceptable = !rec.text.nil?
-      else
-        rec.text = source_advice
-        party_short_code = rec.party_short_code_from_text
-        acceptable = !party_short_code.nil?
-      end
+      # ------------------------------------------------------------------------
 
-      if rec.text == "" || !acceptable
+      canonical_advice = source_advice.downcase.parameterize(separator: "_").to_sym
```
aspiers, line 46: Seems fine although I'm somewhat curious why you just didn't take `source_advice` unmodified as the keys for `SMV_CODES_BY_RECOMMENDATION_TEXT`.

db/fixtures/tactical_vote_stt_recs.rb
```diff
@@ -31,31 +41,48 @@ def load
...
+      party_smv_code = SMV_CODES_BY_RECOMMENDATION_TEXT[canonical_advice]
+      non_party_advice = if party_smv_code.nil?
+                           ACCEPTABLE_NON_PARTY_ADVICE.include?(canonical_advice) ? canonical_advice : nil
+                          else
+                            nil
+                          end
+
+      if party_smv_code && parties_by_smv_code[party_smv_code]
+        rec.text = party_smv_code.to_s.titleize
```
aspiers, line 51: Why not use `party.name` from the `Party` model?  Wouldn't that ensure that party names are displayed consistently everywhere?

aspiers (approved):

app/models/party.rb
```diff
@@ -26,10 +26,10 @@ class Party < ApplicationRecord
 
   def standing_in(ons_id)
     if smv_code.to_sym == :plaid
-      return ons_id[0] == 'W'
+      return ons_id[0] == "W"
```
aspiers, line 29: Would be nice to squash these fixes into the originating commit within the PR in future, to keep the commit history clean, avoid tests failing on certain commits, and keep the possibility of using `git bisect` to find other issues.

baob: I try not to get to this situation, but I missed these locally (maybe they're missed by our rubocop/guard combo, unless you see it at the time of the actual commit), but these only surfaced after the PR was created.  So I tend to simply add commits to an existing PR just so it's clear what's happened, because I lean towards not doing anything that would need a force push after the PR was created, because it's not always obvious why a force-push has happened, and it can get messy if someone has already pulled the branch.

Next time I hit this I will try to fix similar things quickly after the PR as raised.

db/fixtures/tactical_vote_stt_recs.rb
baob: The text property of the Recommendation is presently displayed in the recommendations table which does not have much horizontal space, hence the use of (mostly) 3 letter codes.

Regarding uniqueness, using the party name creates an unwelcome linkage between UI and code. If someone should decide the name shown on the UI should be 'Conservative' instead of 'Conservatives' then any code that refers to 'Conservatives' has to change.

db/fixtures/tactical_vote_stt_recs.rb
baob: I could have done, but some of these imports are inconsistent in case, which would lead to more entries in the lookup. Also, a lookup hash with symbols rather than double-quoted text always looks cleaner to me.

Just preferences 🤷

db/seeds_for_general_election.rb
baob: These tables only ever change at seed time which might be
* initialising database
* re-runs for updated party list, or new recommendations, etc.
So the checks would catch any problems

My gut feeling (and vague recollection) is that adding foreign key constraints to tables accessed via activeRecord  only leads to problems. You can find yourself fixing problems that only exist because of the constraints. (Could be that I'm thinking of rails before activeRecord supported foreign keys though.

On balance I thought these checks were a better way to go than foreign key constraints.

aspiers: > These tables only ever change at seed time which might be
> 
> * initialising database
> * re-runs for updated party list, or new recommendations, etc.
>   So the checks would catch any problems

Sure - to be clear, it's great to have this check and I'm certainly not arguing against having it :)

> My gut feeling (and vague recollection) is that adding foreign key constraints to tables accessed via activeRecord only leads to problems. You can find yourself fixing problems that only exist because of the constraints. (Could be that I'm thinking of rails before activeRecord supported foreign keys though.
> 
> On balance I thought these checks were a better way to go than foreign key constraints.

I can imagine operations like deletion via activeRecord possibly running into problems if it didn't handle foreign keys correctly (and funnily enough see [this ancient answer I wrote](https://stackoverflow.com/a/7805719/179332) which must be hopelessly out of date by now).  However, we're not talking about foreign keys here, just a uniqueness constraint within the table.

db/fixtures/tactical_vote_stt_recs.rb
aspiers: Thanks for the explanation!

aspiers:
BTW I forgot to say thanks for the excellent PR description - super helpful when reviewing!!

app/models/party.rb
aspiers: No worries.  Personally [I prefer a force-push](https://blog.adamspiers.org/2017/08/16/squash-merging-and-other-problems-with-github/) because even though [GitHub](https://github.com/isaacs/github/issues/997) [miserably](https://github.com/isaacs/github/issues/998) [fails](https://github.com/isaacs/github/issues/999) to provide a good UI for this, it's just about usable enough.  But this is down to personal preference really.

db/migrate/20240607094541_add_smv_code_to_party.rb
```diff
@@ -0,0 +1,20 @@
+class AddSmvCodeToParty < ActiveRecord::Migration[5.2]
+  def up
+    add_column :parties, :smv_code, :string
```
baob, line 3: Unique SMV party code created, begins to address #458

Issue #763, opened 2024-06-01
BUG: test database setup - can make developer (rspec) testing impossible


baob:
There's no simple way to re-create the test database. I discovered this after accidentally causing huge numbers of test to fail by running seeds against the test DB.

What should work, by convention, is

    RAILS_ENV=test rake db:seed
    
Actually, after that lots of tests fail and you need to spot what tests are failing due to unexpected extra table rows in the test DB, and then empty  those tables. From memory, emptying 3 or 4 tables this way will do it.

@aspiers you're probably the only one affected by this right now.

aspiers:
What's the cause of this, and what's your workaround?

baob:
> What's the cause of this, and what's your workaround?

The trigger for this could be anything that writes unwanted table rows to the test db. Like developer error. May also affect someone setting up and new test DB.

Underlying cause is probably that we have seed scripts that don't take into account the differing needs for for test. E.g. for our current suite of tests it appears some tables must be empty in the test environment

Workaround is in the first comment (description)

> after that lots of tests fail and you need to spot what tests are failing due to unexpected extra table rows in the test DB, and then empty those tables.

baob:
I may have misunderstood the conventions -> https://stackoverflow.com/questions/55270463/how-to-automatically-migrate-the-development-and-test-databases-with-rails-dbm

baob:
Closed by #777

PR #777, opened 2024-06-09
issue 763 test database setup

baob:
Closes issue #763

- correct link to github readme document
- notes about db migrations and the test db
- add notes on database seeding

### DIFF
--- a/app/views/static_pages/about.html.haml
+++ b/app/views/static_pages/about.html.haml
@@ -83,6 +83,6 @@
       %li
         Developers are welcome to join the project.  The
 
-        = link_to "README document"
+        = link_to "README document", "https://github.com/swapmyvote/swapmyvote/blob/master/README.md"
 
         is the best starting point.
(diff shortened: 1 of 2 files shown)


app/views/static_pages/about.html.haml
```diff
@@ -83,6 +83,6 @@
       %li
         Developers are welcome to join the project.  The
 
-        = link_to "README document"
+        = link_to "README document", "https://github.com/swapmyvote/swapmyvote/blob/master/README.md"
```
aspiers, line 86: Wow good catch, @tdg do we have a Google Search Console or similar set up to catch broken links?

aspiers (approved):

app/views/static_pages/about.html.haml
tdg: I don't think we do, no. I will have a look into it tomorrow

tdg: haha, precisely a year later, this is now set up

aspiers: Yes but also it's possible to set up CI to auto-scan for broken links. I've created #1032 to track this.

PR #788, opened 2024-06-09
add `ELECTION_DATE` and `ELECTION_TYPE` to `.env.example`

aspiers:
These were added to `app.json` but not `.env.example`.


baob (approved):
Good catch !

PR #786, opened 2024-06-09
Improve UI consistency

stevebaxter:
This makes a series of small tweaks to the UI to improve consistency and layout:

1. Remove unwanted padding so panels are displayed correctly
2. Increase narrow container width slightly so text flows better
3. Replace `.text-error` with Bootstrap alerts
4. Show errors next to the error field where possible, or just above the commit button if not

PR #790, opened 2024-06-09
Upgrade to Rails 6.1.7.8 and Ruby 3.3.2

### LINKED ISSUE #778: Switch to newer Ruby version
We're currently on 2.7.4 which is pretty ancient: https://www.ruby-lang.org/en/downloads/releases/

Heroku is screaming at us to upgrade:
image

Also probably we get all kinds of benefits from switching to 3.3.x, including performance.  That said, we have more urgent things to do currently, so we wouldn't want to do this straight away, and certainly not unless it's minimally disruptive.

### LINKED ISSUE #781: Upgrade to Rails >= 6.1.x
We're currently on Rails 5.2.6 which is EOL: https://guides.rubyonrails.org/maintenance_policy.html

Even to get fixes for severe security issues, we need to be on 6.1.x.

So this is pretty urgent.

See also #780 and #778.

aspiers:
Fixes #781 and fixes #778.


aspiers:
@baob Just rebased this to fix conflict with #786.  Have you had a chance to try it yet, and would you be OK with me merging this soon?

aspiers:
Oops, fixing test failure

baob (approved):

PR #791, opened 2024-06-10
issue 773 show aligned tv sites

baob:
Closes #773 and #230

Shows tactical voting recomended sites as sentences on swaps listing

Several phases

# Phase 1 - basics - show good swaps

- basics of showing the new ticks recommendations
- correct population of links on recommendation party table
- add link to recommendation party
- show the original recommendation text
- remove ambiguity about recommendation constituency

# Phase 2 - now show bad swaps

- now also showing bad swaps
- remove the tabular display of recommendations
- reduce site list in helpers/recommendations

# Phase 3 - show recommedation sites that have no recommendations

- add unknowns to recommendations
- some party names fully resolved

# Phase 4 - refactor, normalise data, always use full party name

- document: the plan
- make recommendation site more like AR model
- add basics to RecommendationSite for migration
- quick fix for NI issue - parties become text only
- align the TV imports, look for refactoring opportunities
- use full party names where we can
- passing all normalised data to view
- passing all normalised data to view - part 2
- dont populate the fields we dont need right now

PR #807, opened 2024-06-13
integrate siu tv recs

baob:
Integrate scotland in union tactical vote recommendations

Built on the branch for pull request #791, best review that first

- scotland in union tactical vote csv
- import scripts done
- add scotland in union to the campaigns displayed
- make user seeding a bit more robust
- only show scotland in union on scottish constituencies

image


tdg:
Re 'only show scotland in union on scottish constituencies' I wonder if it isn't quite a good idea to show 'no recommendation' instead. Display our non-partisanship for all to see. 

Speaking of which it'd be interesting - could actually make sense - if they did make recommendations south of the border!

baob:
If that's what you want, fine, to me it didn't make sense, but I can back out that change.

aspiers:
Rebased for clarity of git history.

aspiers (approved):

Issue #675, opened 2022-08-12
optimise speed of db/seed scripts


baob:
At least one issue:
- [ ] Presently iterates twice over polls data (electoral calculus). Once to insert, second to calculate marginal score

aspiers:
Just doing `db:setup` and it's horrifically slow (like 10-30 mins or perhaps even more)!

baob:
I'm not seeing the same thing here:

```
real	3m24.499s
user	2m16.408s
sys	0m49.157s
```
```
$ sqlite3 --version
3.32.3 2020-06-18 14:16:19 02c344aceaea0d177dd42e62c8541e3cab4a26c757ba33b3a31a43ccc7d4aapl
```
How does that compare to you ? Are you on sqlite ?

aspiers:
Did you run `rake db:setup`?  For me that took possibly over an hour.

Yes, on sqlite 3.44.2.

aspiers:
Seems a lot quicker with Rails 5 / Ruby 2.7.6 and #789 but still ~10 minutes or so.

baob:
> Did you run `rake db:setup`?

Yes. Full test below with versions

On current swapmyvote/master
Commit eee48bce0513e937e7f294685284857c8fe06597

```
$ ruby -v
ruby 2.7.6p219 (2022-04-12 revision c9c2245c0a) [x86_64-darwin17]

$ sqlite3 --version
3.32.3 2020-06-18 14:16:19 02c344aceaea0d177dd42e62c8541e3cab4a26c757ba33b3a31a43ccc7d4aapl

$ rails -v
Rails 5.2.6

$ uname -mprsv
Darwin 20.6.0 Darwin Kernel Version 20.6.0: Thu Jul  6 22:12:47 PDT 2023; root:xnu-7195.141.49.702.12~1/RELEASE_X86_64 x86_64 i386

$ rake db:drop

$ time rake db:setup

real	4m12.173s
user	2m46.050s
sys	1m0.496s
```

baob:
Thoughts:
- bug in your version of sqlite3 ?
- missing index (possibly due to above)

baob:
I can improve on that by doing the marginal score thing at at the same time as polls load, but it really looks like you have something else going on.

aspiers:
[...]

PR #829, opened 2024-06-20
upsert all rows of some seed tables in one go

aspiers:
This can save several minutes in the `db:seed` part of `db:setup`.

Partially addresses #675.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -312,8 +312,12 @@ class User < ApplicationRecord
   # This is used to determine whether to enforce the requirement for
   # mobile verification.
   def mobile_verification_missing?
-    return false if test_user? && ENV["TEST_USERS_SKIP_MOBILE_VERIFICATION"]
+    return false if test_user_validation_bypassed?
     return !mobile_phone_verified?
+  end
+
+  def test_user_validation_bypassed?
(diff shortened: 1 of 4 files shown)


aspiers:
@baob This makes seeding stages near instant.

aspiers:
I had a look at trying to do the same for STT recommendations, but it wasn't simple because it relies on `update_parties` in the model and also can do deletes in some scenarios.

aspiers:
But also @baob I see that `Recommendation.update_parties` causes three `SELECT` statements for each insert, which makes it SUPER slow.

baob:
Your upsert code looks good, but the whole seeding process looks broken, almost certainly by the recent password change for users (my change). Are you seeing errors like this on seeding users ? (db:seed:development:users)

> User robert.2-4-1@example.com not created: ["Password can't be blank"]

aspiers:
Yes I am seeing that too.  We need to set a password for test users.

baob:
I've pushed a fixed branch to my repo, but my test setup is messed up, so I can't be sure wether anything is broken. @aspiers could you please cherry pick that fix into your branch ?  My branch has the same name.

aspiers:
@baob Rebased, cherry-picked, and added another commit I forgot to push before.

baob:
Tests still broken for me but I guess they're passing for you and CI.

PR #836, opened 2024-06-21
remove vendored yarn

aspiers:
Not sure if this is a good idea but Heroku seems to recommend it at least:

```
-----> Installing binaries
       engines.node (package.json):   >=16.10 <17
       engines.npm (package.json):    unspecified (use default)
       engines.yarn (package.json):   unspecified (use default)
       packageManager (package.json): yarn@3.8.1
       
        !     Yarn release script may conflict with "packageManager"
       
              The package.json file indicates the target version of Yarn to install with:
              - "packageManager": "yarn@3.8.1"
       
              But the .yarnrc.yml configuration indicates a vendored release of Yarn should be used with:
              - yarnPath: ".yarn/releases/yarn-3.8.1.cjs"
       
              This will cause the buildpack to install yarn@3.8.1 but, when running Yarn commands, the vendored release
              at ".yarn/releases/yarn-3.8.1.cjs" will be executed instead.
       
              To ensure we install the version of Yarn you want, choose only one of the following actions:
              - Remove the "packageManager" field from package.json
              - Remove the "yarnPath" configuration from .yarnrc.yml and delete the vendored release at ".yarn/releases/yarn-3.8.1.cjs"
              https://devcenter.heroku.com/articles/nodejs-support
```


aspiers:
OK, well CI needs a tweak at least:

image

stevebaxter:
Superseded by #1087, which removes the vendored release on current master and adds `corepack enable` to CI to fix the failure seen here.

PR #837, opened 2024-06-21
app.json: update stack to `heroku-22`

aspiers:
This reflects what we are running in staging and will shortly be running in production.  It should also hopefully fix review apps in the pipeline.


aspiers:
This does indeed fix deployment of PR review apps on Heroku, which is a step forward, so I'm merging this - although those review apps aren't yet working.

PR #838, opened 2024-06-21
seed Heroku review apps with a copy of the staging DB

aspiers:
This trick was found in:

https://stackoverflow.com/a/36784098/179332

Hopefully this will get Heroku review apps working.


aspiers:
This isn't working yet.

baob (approved):

PR #843, opened 2024-06-21
update admin guide and db maintenance queries

aspiers:


aspiers:
@baob Is this an example of one of those random test failures with duplicate email?

baob (approved):

baob:
> random test failures with duplicate email?

I've not seen this one before, but the error originates in the prostgres implementation of unique constraints, so not to surprising, as i only use sqlite locally.

Not easy to spot how this error happened.

aspiers:
I guess it's related to rspec's random ordering of tests, because it vanished when I reran the job.  If so, in theory it's reproducible via:

    rspec --seed 19830

but that would require Postgres of course.

It comes from:

https://github.com/swapmyvote/swapmyvote/blob/93bf8d42cd2908d34e18c24e1ce08319030f948f/spec/controllers/users_controller_spec.rb#L31

which refers to a test user in the factory:

https://github.com/swapmyvote/swapmyvote/blob/93bf8d42cd2908d34e18c24e1ce08319030f948f/spec/factories/user.rb#L14-L15

So somehow that user must have already been created somehow.  Presumably not from seeds, since then that test would always fail on the `create`.  So from another test?  Could it have come from here maybe?

https://github.com/swapmyvote/swapmyvote/blob/93bf8d42cd2908d34e18c24e1ce08319030f948f/spec/factories/swaps.rb#L5

PR #848, opened 2024-06-23
set default env variables correctly for user seeding

### LINKED ISSUE #846: rake db:seed fails with "Password can't be blank"
Re-run with rake db:seed:development:users

baob:
Closes #846. Without this setting, seeding the database with users will fail

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -43,7 +43,7 @@ export SWAPMYVOTE_MODE=open
 
 # Uncomment this to allow test users (example.com / Facebook test
 # users) to skip mobile verification
-# export TEST_USERS_SKIP_MOBILE_VERIFICATION=yes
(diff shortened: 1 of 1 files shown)


aspiers (approved):
Approving because this is an important fix, however it feels to me that we still have a problem here: `TEST_USERS_SKIP_MOBILE_VERIFICATION` is being used not only to skip mobile verification but also to determine whether a password is required:

https://github.com/swapmyvote/swapmyvote/blob/f32fee8b7ba4919baed30f8b73ee8ae8305665f5/app/models/user.rb#L370-L371

This definitely seems wrong to me.  I don't think we should ever have empty passwords; we can just seed test users with a fixed password like `swapmyvote` or a random string.  (If the latter, the unencrypted random string could be output during seeding, to allow capture for later use.)  Filed as #852.

PR #847, opened 2024-06-23
document the need to copy test environment example

baob:
Closes #845. Without this step 5 unit tests fail locally.

PR #853, opened 2024-06-23
Add party labels to prediction chart

### LINKED ISSUE #825: label parties on bar chart
For me a large part of the struggle with the bar chart is that the parties aren't labelled (only visible via mouse hover), and I haven't perfectly internalised the various party colours.  It should be pretty easy to fix this, e.g. via one or both of:

- Include the party logo in the bar.
- Include the party name in the bar (rotated vertically to fit, using CSS).

_Originally posted by @aspiers in https://github.com/swapmyvote/swapmyvote/issues/816#issuecomment-2179206246_

stevebaxter:
This adds a short name to the prediction chart.

Closes #825.

### DIFF
--- a/app/assets/javascripts/polls.coffee
+++ b/app/assets/javascripts/polls.coffee
@@ -10,7 +10,6 @@ _drawCharts = () ->
 google.charts.load('1', {packages: ['corechart']})
 google.setOnLoadCallback(_drawCharts)
 
-
 _drawPollChart = (selector, poll_data) ->
   formatMS = new google.visualization.NumberFormat({
     suffix: '%',
@@ -26,28 +25,27 @@ _drawPollChart = (selector, poll_data) ->
   max = Math.max.apply(null, poll_data.map((d) -> d[1]))
 
   options = {
-    hAxis: { textPosition: 'none' },
+    hAxis: {
+      textPosition: 'out'
+    },
     vAxis: {
       textPosition: 'none',
       gridlines: {
         color: 'transparent'
       },
       viewWindow: {
-        min: 0, max: max
+        min: 0,
+        max: max
       },
       baselineColor: 'none',
     },
     legend: { position: "none" },
     bar: { groupWidth: '90%' },
     chartArea: {
-      width: '100%',
-      height: '100%'
+      width: '100%'
     },
     tooltip: {
-      textStyle: {
-        fontSize: 12,
-        fontFamily: "Open Sans"
-      }
+      trigger: "none"
     }
   }
 
--- a/app/helpers/polls_helper.rb
+++ b/app/helpers/polls_helper.rb
@@ -2,7 +2,7 @@ module PollsHelper
   def poll_data_for(constituency)
     data = []
     constituency.polls.sort_by {|p| p.votes}.reverse.each do |poll|
-      data.push [poll.party.name, poll.votes / 100, poll.party.color] unless poll.votes.zero?
+      data.push [poll.party.short_name, poll.votes / 100, poll.party.color] unless poll.votes.zero?
     end
     return data.to_json.html_safe
   end
--- a/app/models/party.rb
+++ b/app/models/party.rb
@@ -5,19 +5,19 @@ class Party < ApplicationRecord
            dependent: :destroy
 
   REFERENCE_DATA = {
-    lab: { name: "Labour", color: "#DC241f", ge_default: true },
-    libdem: { name: "Liberal Democrats", color: "#FFB602", ge_default: true },
-    green: { name: "Green", color: "#6AB023", ge_default: true },
-    con: { name: "Conservatives", color: "#0087DC", ge_default: true },
-    ukip: { name: "UKIP", color: "#70147A" },
-    snp: { name: "SNP", color: "#FFF95D", ge_default: true },
-    plaid: { name: "Plaid Cymru", color: "#008142", ge_default: true },
-    brexit: { name: "Brexit", color: "#5bc0de" },
+    lab: { name: "Labour", color: "#DC241f", short_name: "Lab", ge_default: true },
+    libdem: { name: "Liberal Democrats", short_name: "LD", color: "#FFB602", ge_default: true },
+    green: { name: "Green", color: "#6AB023", short_name: "Green", ge_default: true },
+    con: { name: "Conservatives", color: "#0087DC", short_name: "Con", ge_default: true },
+    ukip: { name: "UKIP", short_name: "UKIP", color: "#70147A" },
+    snp: { name: "SNP", short_name: "SNP", color: "#FFF95D", ge_default: true },
+    plaid: { name: "Plaid Cymru", short_name: "PC", color: "#008142", ge_default: true },
+    brexit: { name: "Brexit", short_name: "Brexit", color: "#5bc0de" },
     breakthrough: { name: "Breakthrough" },
     for_britain: { name: "For Britain" },
     freedom_alliance: { name: "Freedom Alliance" },
-    independent: { name: "Independent" },
-    reform: { name: "Reform", color: "#5bc0de", ge_default: true },
+    independent: { name: "Independent", short_name: "Ind" },
+    reform: { name: "Reform", short_name: "Ref", color: "#5bc0de", ge_default: true },
     rejoin_eu: { name: "Rejoin eu" },
     workers: { name: "Workers" },
     yorkshire: { name: "Yorkshire" },
@@ -31,6 +31,12 @@ class Party < ApplicationRecord
       return ons_id[0] == "S"
     end
     return true
+  end
+
+  def short_name
+    short = smv_code ? REFERENCE_DATA[smv_code.to_sym][:short_name] : nil
+    short ||= name.downcase.gsub(/[aeiou]/, "")[0..2].titleize
+    short
   end
 
   class << self
--- a/spec/views/layouts/__snapshots__/application_by_election.snap
+++ b/spec/views/layouts/__snapshots__/application_by_election.snap
@@ -25,7 +25,7 @@
 <link href='https://stackpath.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css' rel='stylesheet'>
 <script src='https://www.gstatic.com/charts/loader.js' type='text/javascript'></script>
 <link rel="stylesheet" media="all" href="/assets/application-4ecdfd9767b8e36e6bd5084595095b0c5796771f9d39d730fd286ebe33691303.css" data-turbolinks-track="true" />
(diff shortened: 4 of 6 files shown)


app/assets/javascripts/polls.coffee
```diff
@@ -26,22 +25,24 @@ _drawPollChart = (selector, poll_data) ->
   max = Math.max.apply(null, poll_data.map((d) -> d[1]))
 
   options = {
-    hAxis: { textPosition: 'none' },
+    hAxis: {
+      textPosition: 'out'
```
stevebaxter, line 29: Show H axis labels

app/assets/javascripts/polls.coffee
```diff
@@ -26,22 +25,24 @@ _drawPollChart = (selector, poll_data) ->
...
+        max: max
       },
       baselineColor: 'none',
     },
     legend: { position: "none" },
     bar: { groupWidth: '90%' },
     chartArea: {
-      width: '100%',
-      height: '100%'
+      width: '100%'
```
stevebaxter, line 45: We need to change this to allow enough space for the labels

app/helpers/polls_helper.rb
```diff
@@ -2,7 +2,7 @@ module PollsHelper
   def poll_data_for(constituency)
     data = []
     constituency.polls.sort_by {|p| p.votes}.reverse.each do |poll|
-      data.push [poll.party.name, poll.votes / 100, poll.party.color] unless poll.votes.zero?
+      data.push [poll.party.shortName, poll.votes / 100, poll.party.color] unless poll.votes.zero?
```
stevebaxter, line 5: Use the short name here.

app/assets/javascripts/polls.coffee
```diff
@@ -26,28 +25,27 @@ _drawPollChart = (selector, poll_data) ->
...
     },
     tooltip: {
-      textStyle: {
-        fontSize: 12,
-        fontFamily: "Open Sans"
-      }
+      trigger: "none"
```
stevebaxter, line 48: Disable tooltips now all the data is on the chart

app/models/party.rb
```diff
@@ -5,19 +5,19 @@ class Party < ApplicationRecord
...
   REFERENCE_DATA = {
-    lab: { name: "Labour", color: "#DC241f", ge_default: true },
-    libdem: { name: "Liberal Democrats", color: "#FFB602", ge_default: true },
-    green: { name: "Green", color: "#6AB023", ge_default: true },
-    con: { name: "Conservatives", color: "#0087DC", ge_default: true },
-    ukip: { name: "UKIP", color: "#70147A" },
-    snp: { name: "SNP", color: "#FFF95D", ge_default: true },
-    plaid: { name: "Plaid Cymru", color: "#008142", ge_default: true },
-    brexit: { name: "Brexit", color: "#5bc0de" },
+    lab: { name: "Labour", color: "#DC241f", shortName: "Lab", ge_default: true },
```
stevebaxter, line 8: Add to the reference data

db/schema.rb
```diff
@@ -48,6 +48,7 @@
   create_table "parties", force: :cascade do |t|
     t.string "name"
     t.string "color"
+    t.string "shortName"
```
stevebaxter, line 51: I **think** this is the right thing to do? I don't know anything about how Ruby does databases.

db/seeds/development/users.seeds.rb
```diff
@@ -12,7 +12,8 @@ def create_random_user(i, preferred_party_id, willing_party_id)
     email: "#{firstname.downcase}.#{i}-#{preferred_party_id}-#{willing_party_id}@example.com",
     constituency_ons_id: ons_id,
     preferred_party_id: preferred_party_id,
-    willing_party_id: willing_party_id
+    willing_party_id: willing_party_id,
+    password: "123456"
```
stevebaxter, line 16: I think we need this now passwords are required. Without this we don't get any users when we seed the DB.

db/seeds_for_by_election.rb
```diff
@@ -18,7 +18,7 @@ class SeedHelper
 puts "\nParties"
 
 Db::Fixtures::Be2022::Party.all.each do |party|
-  ::Party.find_or_create_by(name: party[:name], color: party[:colour])
+  ::Party.find_or_create_by(name: party[:name], shortName: party[:shortName], color: party[:colour])
```
stevebaxter, line 21: I **think** this is the right thing to do? I don't know anything about how Ruby does databases.

db/seeds_for_general_election.rb
```diff
@@ -24,7 +24,7 @@
 
 parties_for_ge.each do |party_attributes|
   party = Party.find_or_initialize_by(name: party_attributes[:name])
-  party.update!(party_attributes.slice(:smv_code, :color))
+  party.update!(party_attributes.slice(:smv_code, :color, :shortName))
```
stevebaxter, line 27: I **think** this is the right thing to do? I don't know anything about how Ruby does databases.

spec/views/layouts/__snapshots__/application_by_election.snap
```diff
@@ -42,7 +42,7 @@
       },
       cookie: {
         name: "_swapmyvote_cookie_consent",
-        domain: "localhost"
+        domain: ""
```
baob, line 45: There's a fix for this, this part of the snapshot should not be changing. 

You need to create an `env.test.local` .   See https://github.com/swapmyvote/swapmyvote/pull/847/files

spec/views/layouts/__snapshots__/application_general.snap
```diff
@@ -42,7 +42,7 @@
       },
       cookie: {
         name: "_swapmyvote_cookie_consent",
-        domain: "localhost"
+        domain: ""
```
baob, line 45: There's a fix for this, this part of the snapshot should not be changing. 

You need to create an `env.test.local` .   See https://github.com/swapmyvote/swapmyvote/pull/847/files

baob (changes requested):

db/schema.rb
aspiers: It is but I guess we need to add a migration, and also for consistency we should stick to the `snake_case` convention rather than `camelCase`.

db/seeds_for_general_election.rb
baob: It's not as simple as that to add it to the database. What about just keeping it in the model ?

spec/views/layouts/__snapshots__/application_general.snap
stevebaxter: I've changed it back

stevebaxter:
OK, I think everything is passing now.

stevebaxter:
It looks like this now:

Screenshot 2024-06-23 at 18 06 51

The tooltips have been removed because they don't contain any extra data now.

aspiers (changes requested):
Looks awesome!  Just missing the db migration as discussed, but maybe I can find time to do this while you're away.

db/schema.rb
```diff
@@ -48,6 +48,7 @@
   create_table "parties", force: :cascade do |t|
     t.string "name"
     t.string "color"
+    t.string "short_name"
```
aspiers, line 51: I don't think we need this as per Zulip discussion; instead let's define a `#short_name` method on the model to do the lookup against `REFERENCE_DATA`.

baob:
I added two commits to add the short_name function to my own copy of that branch. Use or ignore as you want :)  
https://github.com/baob/swapmyvote/commits/feature/add-party-labels/

stevebaxter:
@baob - let's use your branch!

aspiers:
@baob Please can you submit a PR?

aspiers:
Actually no, let's just update this one.

aspiers:
It's a bit nicer to have a single PR otherwise the review history gets fragmented across two PRs.  We can just force-push this one to point to Andy's branch.

stevebaxter:
Is there anything I need to do here? I'm not sure how to force-push this across forks?

baob:
> Is there anything I need to do here?

This is what I'd do if sitting at your machine

```
git remote add baob git@github.com:baob/swapmyvote.git
git fetch baob
git checkout feature/add-party-labels 
git cherry-pick baob/feature/add-party-labels~1
git cherry-pick baob/feature/add-party-labels 
```

Not quite what @aspiers suggested but that should work.

stevebaxter:
Hmm, no dice unfortunately:
```
stevebaxter@StevesMiniM2Pro swapmyvote % git remote add baob git@github.com:baob/swapmyvote.git
stevebaxter@StevesMiniM2Pro swapmyvote % git fetch baob
The authenticity of host 'github.com (20.26.156.215)' can't be established.
ED25519 key fingerprint is SHA256:+DiY3wvvV6TuJJhbpZisF/zLDA0zPMSvHdkr4UvCOqU.
This key is not known by any other names.
Are you sure you want to continue connecting (yes/no/[fingerprint])? yes
Warning: Permanently added 'github.com' (ED25519) to the list of known hosts.
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
stevebaxter@StevesMiniM2Pro swapmyvote % git checkout feature/add-party-labels 
Already on 'feature/add-party-labels'
Your branch is up to date with 'origin/feature/add-party-labels'.
stevebaxter@StevesMiniM2Pro swapmyvote % git cherry-pick baob/feature/add-party-labels~1
fatal: bad revision 'baob/feature/add-party-labels~1'
stevebaxter@StevesMiniM2Pro swapmyvote % git cherry-pick baob/feature/add-party-labels  
fatal: bad revision 'baob/feature/add-party-labels'
```
Honestly I would just open a PR from your branch. It's going to be much quicker than messing with git.

aspiers (approved):

aspiers:
I've taken care of this.

@stevebaxter This was your problem:
> git@github.com: Permission denied (publickey).

Something's quite wrong with your setup for it not to be able to pull from a public read-only github repo via ssh.  I suspect you'll have the same problems fetching from other github repos via ssh.  The following seems like a hint:

> The authenticity of host 'github.com (20.26.156.215)' can't be established.
> ED25519 key fingerprint is SHA256:+DiY3wvvV6TuJJhbpZisF/zLDA0zPMSvHdkr4UvCOqU.
> This key is not known by any other names.

This suggest to me that either you've never accessed github over ssh before and only used https, or something like that?  ssh is the recommended transport these days.

In any case, given that fetching from Andy's remote failed, it was inevitable that cherry-picking commits from his remote would fail too.  BTW cherry-picking is not needed, it was enough to do `git reset --hard baob/feature/add-party-labels` - but that does also depend on first doing a successful fetch.

I also squashed the three commits into one before merging, as the last two were just tidying up things in the first, so the history is cleaner this way.  It looks good on my machine - great work both!!

image

stevebaxter:
I only ever use GH Desktop, so only ever use https. Thanks for merging!

PR #860, opened 2024-06-25
issue 816 3-way polling advice text

### LINKED ISSUE #816: FEATURE: TV annotation win/lose/make a difference, based on polls
Annotate a single TV recommendation (so, we know the constituency and the party that's going to get the vote in that constituency)

Feature is variously described. By @baob 

> inferences from polling data (losing/making a difference/winning) ?

By @aspiers 

> making the bar graph easier to understand, e.g. hiding it behind some text using maybe three categories of descriptions:
> * your preferred party is polling a lead of X% in this constituency,
> * polls say it's marginal / close call,
> * your preferred party is 2nd in the polls in this constituency, trailing by Y% (although it feels like we also need:
> * your preferred party is 3rd / 4th / ...

For first cut, I'm going to aim for basing this on the marginal score and (new) signed marginal score
* "(green tick) polling suggests your swap could make a difference for this party"
  - because marginal score < 10%, they're likely to just win or just lose
* "polling suggests your swap is supporting a safe win for this party"
   - because signed marginal score > 10%, positive they are way ahead
* "polling suggests your party will still lose despite your swap"
   - because signed marginal score < 10% negative, they are way behind

"this party" will probably be the actual party name. 

/cc @tdg

baob:
Closes #816 following extended discussion there

- **optimise marginal score calculation**
- **fix tests broken by eager load**
- **refactor, extract new method**
- **basics of 3-way polling advice**
- **simple advice, no other parties**
- **add percentages but no other parties**
- **move the polls interpretation panel closer to polls chart**
- **don't need check the polls now**
- **use percent sign**
- **remove polls interpretation**
- **rounding percent**
- **change some wording**
- **star not bang to highlight marginal**

image

### DIFF
--- a/app/models/poll.rb
+++ b/app/models/poll.rb
@@ -29,13 +29,16 @@ class Poll < ApplicationRecord
       # rubocop:enable Rails/FindEach
         polls = constituency.polls
         polls.each do |poll|
-          party_votes = poll.votes
-          max_votes = polls.select { |p| p.id != poll.id }.map(&:votes).max
-          poll.update(marginal_score: (max_votes - party_votes).abs)
+          poll.update(marginal_score: (poll.signed_marginal_score).abs)
           print "." if progress
         end
       end
       puts if progress
     end
+  end
+
+  def signed_marginal_score
+    other_poll_votes = constituency.polls.select { |p| p.id != id }.map(&:votes)
+    votes - other_poll_votes.max
   end
 end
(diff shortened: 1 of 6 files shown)


stevebaxter:
That looks excellent!

stevebaxter (approved):
LGTM

baob:
@aspiers Implemented your rounding suggestion, now looks like this 

image

baob:
@aspiers we should probably get this merged and onto the testing site, too.

aspiers:
I know, already reviewing!

aspiers (approved):

PR #862, opened 2024-06-25
Improve test user seeding

### LINKED ISSUE #852: never allow empty passwords
As noted in https://github.com/swapmyvote/swapmyvote/pull/848#pullrequestreview-2134297153, when `TEST_USERS_SKIP_MOBILE_VERIFICATION` is set to `yes`, we allow empty passwords since [`a287a9a` (#829)](https://github.com/swapmyvote/swapmyvote/pull/829/commits/a287a9a9ed0dc50b2bbf9bf4067f7054db07360f#diff-9802ca3c9c4cf89904fd44bc114e35ebdf2c5dd3d5b645491e2b253e1afef29bR371).

However, this leads to an extra usage of `TEST_USERS_SKIP_MOBILE_VERIFICATION` which doesn't match its name; it would need to be renamed to `TEST_USERS_SKIP_MOBILE_AND_PASSWORD_VERIFICATION` which is a huge mouthful.

Instead, I think we can just never allow empty passwords and fix seeding a different way by ensuring that a password is always set, even for test users.  This will provide a staging environment which is closer to production, and also provide a small amount of defence against outsiders messing around with the test users.

aspiers:
Fixes #852.

### DIFF
--- a/db/seeds/development/users.seeds.rb
+++ b/db/seeds/development/users.seeds.rb
@@ -3,16 +3,14 @@ def create_random_user(i, preferred_party_id, willing_party_id)
   gender = rand > 0.5 ? "female" : "male"
   firstname = gender == "male" ? Random.firstname_male : Random.firstname_female
 
-  ons_constituency_id = OnsConstituency.all.pluck(:id).sample
-
-  ons_id = OnsConstituency.find_by(id: ons_constituency_id).ons_id
+  constituency = OnsConstituency.all.sample
   password = firstname + firstname
 
   user = User.create(
     name: "#{firstname} #{Random.lastname}",
     email: "#{firstname.downcase}.#{i}-#{preferred_party_id}-#{willing_party_id}@example.com",
     password: password,
-    constituency_ons_id: ons_id,
+    constituency_ons_id: constituency.ons_id,
     preferred_party_id: preferred_party_id,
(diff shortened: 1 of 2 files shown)


db/seeds/development/users.seeds.rb
```diff
@@ -6,10 +6,12 @@ def create_random_user(i, preferred_party_id, willing_party_id)
   ons_constituency_id = OnsConstituency.all.pluck(:id).sample
 
   ons_id = OnsConstituency.find_by(id: ons_constituency_id).ons_id
+  password = firstname + firstname
```
baob, line 7: 👍 that would have been my choice

db/seeds/development/users.seeds.rb
```diff
@@ -44,24 +44,14 @@ def provider_array
...
-  create_random_user(i, 2, 4)
-
-  create_random_user(i, 4, 3)
-  create_random_user(i, 3, 4)
+  parties.each do |preferred|
+    parties.each do |willing|
+      next if preferred.id == willing.id
+      create_random_user(i, preferred.id, willing.id)
+    end
+  end
```
baob, line 55: Nice. I had this in mind but there was never a pressing need.

baob (approved):

PR #851, opened 2024-06-23
issue 229 tactical campaign names

baob:
Closes issue #229, providing name for all the TV campaigns

- **issue-229: correct formatting**
- **issue-229: add names to TV site data**
- **issue-229: add meta description info for each site**
- **issue-229: replace host names with TV site names on the website**

Demonstrating mouse-over:

image

Issue #229, opened 2019-12-01
Tactical campaigns need renaming


baob:
@tdg has requested:

* could we name the sites with Names like GetVoting, Remain United,
* is it possible easily to swap the order of the two tacticalvote sites so that .co.uk comes first?

I think it makes sense to understand what other supplementary info about the recommendation _sites_ we need to display and work on it together. Obviously from the above we need

* SMV-standardised name
* Sort order 

Do we need a colour code too ?  What about remain versus leave ? 

[This PR](https://github.com/swapmyvote/swapmyvote/pull/227) needs to be merged before this issue can be completed, but it can be started at any time.

aspiers:
Please discuss with @tdg if possible to refine this card into concrete actions. I recommend splitting into multiple issues (e.g. colour and remain vs. leave seem separate to me).

baob:
Some of this info is in the json, just not extracted yet
```
    {
      "siteName": "GetVoting",
      "siteUrl": "https://getvoting.org",
      "site": "get-voting",
      "features": [
        "<strong>Stop Brexit</strong>",
        "By <a href=\"https://www.bestforbritain.org/\" target=\"_new\">Best for Britain</a>",
        "Method <a href=\"https://getvoting.org/faq.html\" target=\"_new\">MRP</a>"
      ]
    },
```

baob:
And @aspiers asked for 

> Actually we should seriously consider displaying "last updated at ..." somewhere

aspiers:
@baob Not relevant for BE2021 but didn't we (well, you) complete this?  Or is it kept open because we didn't implement the "last updated at..." suggestion?

baob:
Natural follow-on when the PR for issue #773 gets merged. It will largely revolve around adding more data to the recommendations helper, then using that in the view.

@aspiers we were talking about this on Sunday afternoon.

baob:
@tgd @aspiers Can you give me an agreed list of 'campaign names' to replace the domain names please ?

I'll make myself busy adding meta descriptions from the campaigns websites for use as mouse hovers, or whatever.

baob:
> * could we name the sites with Names like GetVoting, Remain United,

@aspiers @tdg  Is this still important ?

aspiers:
I wouldn't say _super_ important - presumably presenting the TV campaign names according to their chosen branding is more cosmetic than anything else ... but isn't it also extremely quick easy to do?

tdg:
Stop The Tories 
[...]

PR #863, opened 2024-06-26
tweak TV campaign list

aspiers:
As per https://github.com/swapmyvote/swapmyvote/pull/851#issuecomment-2190041507 following on from #229.


baob (approved):
Changing the order of the entries in the hash (versus changing just the ordering parameter) made it a touch harder to review, but it all works when checked out.

PR #885, opened 2024-06-27
tweak copy on swaps page

aspiers:


tdg (commented):
I actually like the two levels of subjectivity – one for polls, one for elections

baob:
While you're in there the "which is" text is problematic. Works fine for singular party names ("labour") less so for plurals ("conservatives", "liberal democrats").

Extra points if you can squeeze is a simple fix  :) ... otherwise merge as is and tackle this elsewhere.

tdg:
'who are'

tdg:
> I actually like the two levels of subjectivity – one for polls, one for elections

Also, I like how _human_ it sounds - we could do with more of that sense in our copy, which tends to the dry (I should know!)

aspiers:
Dropped the commit removing "Looks like".

baob (approved):

Issue #591, opened 2022-06-09
redesign "Log in" modal flow


aspiers:
One problem with the "Log in" modal (see screenshot below) is that it is used as a precursor for both for registrations and logins to already registered accounts, even though the text only talks about login rather than registration - so that's confusing to the user.

A second, consequent problem is that it is not a good place to ask for consent to process personal info, because we should only do that once per user at registration time (as per #245).

To elaborate: after clicking "Already been here? Log in", there is a checkbox asking for consent:

![image](https://user-images.githubusercontent.com/100738/172889381-8afa8247-308f-40aa-b131-a05c89a09401.png)

This kind of made sense when we had Facebook and Twitter logins enabled, because in that case a) login _was_ effectively registration, and b) there were no further screens on our side as part of the login process, so it made sense to gate login up front on obtaining consent.

But now we don't support Twitter/Facebook logins, only email.  And also we can't record consent in the db until we have an email address or some other form of identity, so it makes more sense to record consent at registration time, on the "Sign up with email" page.  Currently this page only asks for opt in to the newsletter, but it should really ask for consent to process personal info too.

tdg:
NB Consent to process PI is mandatory (but must be understood)
Consent to newsletter must be non-mandatory

In many ways, having them at different parts of the flow can be a feature not a bug, as it means both are likely toi be considered accordingly to the user's intentions.

aspiers:
Sure, to be clear I'm not arguing for removing the ability to opt in to the newsletter at other places.  But:

1. It makes no sense to ask for consent to process PII _every_ time they login.
2. If we can (again) offer sign-up to newsletter at registration, I think we should do it.

Also we can make it clear which is mandatory via the usual red asterisk superscript plus corresponding footnote.

baob:
> 1. It makes no sense to ask for consent to process PII _every_ time they login.

💯

baob:
Does the title "rethink" mean this still needs refinement ?  It doesn't look actionable to me, states a problem, with no direct suggestions for a solution ?

aspiers:
The action is to figure out a solution and then file tasks for the implementation.

aspiers:
@tdg Please can you revisit this and assess urgency for GE2024?

tdg:
Yes it seems urgent, but I don't know what the solution would be @aspiers

baob:
I think the proposed solution is in the first comment isn't it ?

[...]

PR #889, opened 2024-06-27
Change "log in" to "log in or sign up"

stevebaxter:
On the user page we need to make it clear that you can log in or sign up.

Partially addresses #591.

### DIFF
--- a/app/views/layouts/_login.html.haml
+++ b/app/views/layouts/_login.html.haml
@@ -5,7 +5,7 @@
   .modal-dialog
     .modal-content
       .modal-header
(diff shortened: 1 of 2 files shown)


stevebaxter:
Looks like this now:

Screenshot 2024-06-27 at 21 59 42

baob:
Hi @stevebaxter there are some specs to fix here:  spec/views/layouts/_login.html.haml_spec.rb

Should be easy spot what needs to change.

stevebaxter:
Yep, done already!

baob (approved):

PR #893, opened 2024-06-28
Reinstate pageview

egglestn:


aspiers:
Not sure this is necessary because we simply had the wrong analytics id set.  I've fixed that; please can you or @tdg check whether it's working now?  If so we can close this without merging.

egglestn:
@aspiers sadly, no, it's not working yet. I'll put an update in zulip with my best understanding

aspiers (approved):

PR #895, opened 2024-06-28
Add an invisible CAPTCHA

### LINKED ISSUE #861: add invisible_captcha for increased Sybil-resistance
Fortunately we managed to resolve #831 so we have a pretty degree of [Sybil-resistance](https://en.wikipedia.org/wiki/Sybil_attack) to bots etc.  However we could benefit from the option of another line of defence, and https://github.com/markets/invisible_captcha looks like a great option since it is invisible to users (hence the name).

stevebaxter:
We add an invisible CAPTCHA using https://github.com/markets/invisible_captcha. This is added to the email sign-up form.

Closes #861

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -66,6 +66,8 @@ gem "haml_lint"
 
 gem "webpacker", "~> 5.x"
 
+gem "invisible_captcha", "~> 2.3"
+
 # These need to be outside the development group for Rakefile to
 # be happy in Heroku.
 gem "rubocop", "~> 1.22.1"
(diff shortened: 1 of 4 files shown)


aspiers (approved):
Awesome!  Successfully tested by selecting honeypot form field in inspector and then `$0.value = 'I am a bot'` in browser console before submitting.  Resulted in:

    [Invisible Captcha] Potential spam detected for IP ::1. Honeypot param 'registration.qtmgykxnpu-dbwvjia' was present.

PR #897, opened 2024-06-28
Improve sign-up flow

### LINKED ISSUE #591: redesign "Log in" modal flow

stevebaxter:
This removes the "choice" dialog from the flow as we only support email now. We get data processing consent on the email sign-up form now.

We add "required" stars on all the forms that have required fields.

We also disable the timeout on the captcha as it can cause problems when the user clicks submit multiple times.

Closes #591


app/assets/stylesheets/home.scss
```diff
@@ -74,7 +74,7 @@ select {
 }
 
 .container-extra-narrow {
-    max-width: 450px;
+    max-width: 500px;
```
stevebaxter, line 77: Make this a bit wider so the data processing consent checkbox fits

app/controllers/users/registrations_controller.rb
```diff
@@ -1,8 +1,14 @@
...
 
 class Users::RegistrationsController < Devise::RegistrationsController
+  include ApplicationHelper
+
   before_action :configure_sign_up_params, only: [:create]
-  invisible_captcha only: [:create]
+
+  # We disable timestamps because it can cause problems if the user clicks the submit button
+  # several times in a row
+  invisible_captcha only: [:create], timestamp_enabled: false
```
stevebaxter, line 10: I found this was a problem

app/views/devise/passwords/edit.html.erb
```diff
@@ -11,12 +11,12 @@
         <%= form_for(resource, as: resource_name, url: password_path(resource_name), html: { method: :put }) do |f| %>
           <%= f.hidden_field :reset_password_token %>
 
-          <div class="form-group">
+          <div class="form-group required">
```
stevebaxter, line 14: Adding `required` classes to these will add the red asterisk to their labels.

app/views/devise/registrations/new.html.erb
```diff
@@ -2,34 +2,46 @@
...
             <%= f.password_field :password_confirmation, autocomplete: "new-password", class: "form-control" %>
           </div>
 
           <div class="form-group form-check">
             <%= f.check_box :consent_news_email, class: "form-check-input" %>
-            <%= f.label  :consent_news_email, "Opt-in to Forward Democracy email updates", class: "mb-0 form-check-label" %>
+            <%= f.label  :consent_news_email, "Opt-in to Forward Democracy email updates", class: "form-check-label" %>
+          </div>
+
+          <div class="form-group form-check required">
```
stevebaxter, line 35: This is the new checkbox

app/views/home/_swap_form.html.haml
```diff
@@ -40,7 +40,7 @@
     } else if (preferred_party_id == willing_party_id) {
       $(".js-equal-party-error-modal").modal();
     } else {
-      $("#js-login-modal").modal();
+       window.location.assign("#{sign_up_path}")
```
stevebaxter, line 43: Just go straight to the sign up page now.

app/views/layouts/_login.html.haml
```diff
@@ -1,59 +1,2 @@
 %span.user
-  = link_to "Already been here? Log in", "#js-login-modal", class: "small subdued logout", data: {toggle: "modal"}
-.modal#js-login-modal{ tabindex: "-1", role: "dialog",
```
stevebaxter, line 3: I've removed this modal completely as even if we turn FB and Twitter login back on we will need to come back to this flow.

spec/views/layouts/_login.html.haml_spec.rb
```diff
@@ -1,82 +0,0 @@
-require "rails_helper"
```
stevebaxter, line 1: This test is no longer needed because this modal is gone!

baob: Seems we needed that test to keep our coverage up though :D

```
Coverage report generated for RSpec to /home/runner/work/swapmyvote/swapmyvote/coverage. 916 / 1316 LOC (69.6%) covered.
Line coverage (69.60%) is below the expected minimum coverage (70.00%).
Branch coverage ([56](https://github.com/swapmyvote/swapmyvote/actions/runs/9713695862/job/26811036363?pr=897#step:8:57).84%) is below the expected minimum coverage (58.00%).
SimpleCov failed with exit 2 due to a coverage related error
```

baob: I was right, the relevant commit, for code coverage, says "set coverage thresholds so it only just passes".

@stevebaxter please adjust the thresholds a little bit in spec/spec_helper.rb.

PR #919, opened 2024-06-30
issue 905 willing profile

### LINKED ISSUE #905: Willing Party profile
Whenever a voter sets or changes
* the party they are willing to vote for
* their constituency

take then to a page that shows them how their profile will look to potential swappers, and ask them to confirm their choice.

baob:
Closes #905 

Now, when the voter changes either willing party or constituency, they are asked to review their choice and it's impact on the swap profile

- **new method User#swap_profile_changed?**
- **stub action /user/review**
- **refactor action /user/update**
- **update of swap_profile properties redirects to review**
- **fill out review page**
- **add buttons**


app/controllers/users_controller.rb
```diff
@@ -42,6 +45,12 @@ def update
       end
     end
 
+    no_flash_errors = (!flash[:errors] || flash[:errors].size.zero?)
+
+    if @user.valid? && no_flash_errors && review_required
```
aspiers, line 50: Just curious why the user might _not_ be valid here?

app/controllers/users_controller.rb
```diff
@@ -28,9 +28,12 @@ def edit
     @constituencies = OnsConstituency.all.order(:name)
   end
 
+  # rubocop:disable Metrics/MethodLength
   def update
+    review_required = false
     if params[:user]
       @user.assign_attributes(user_params)
+      review_required = @user.swap_profile_changed?
```
aspiers, line 36: Seems quite cunning - I guess this way you don't need a new dirty flag in the DB and just rely on `ActiveRecord::Dirty` instead.

app/views/users/review.haml
```diff
@@ -1,4 +1,19 @@
 = javascript_pack_tag "intlTelInput"
 = stylesheet_pack_tag "intlTelInput"
 
-= render partial: "users/info_summary"
+.plain-pattern.border-bottom
+  .container
+    %p.text-center
+      PLEASE TAKE A MOMENT TO REVIEW THIS.
```
aspiers: Stylistically I'd probably prefer bold to all caps, as the latter feels a bit like shouting - but definitely not a merge blocker, we can debate that later :stuck_out_tongue_winking_eye:

aspiers: Oh, this probably gets changed in subsequent commits anyway...

app/views/user/swaps/_polls_advice_self.html.haml
```diff
@@ -0,0 +1,29 @@
+%p
```
aspiers, line 1: Please can we rename these files so that they don't contain the word "advice" :stuck_out_tongue_winking_eye:

I suggest "interpretation".

baob: Done

app/controllers/users_controller.rb
baob: That's up to the user model to determine. Something left blank ? I was just preserving the original behaviour.

app/controllers/users_controller.rb
baob: Exactly

app/views/users/review.haml
baob: Yes it's deleted later.

aspiers (approved):

app/controllers/users_controller.rb
aspiers: @baob There's a problem with not tracking "willing party has been reviewed and confirmed" in the db, which is that all the existing users without any swaps will never see this page because they've already set their willing.

So I think it's worth having a db field which tracks this to ensure everyone goes through the confirmation of their willing choice.

PR #920, opened 2024-07-01
a gentler introduction to swap matching

### LINKED ISSUE #906: Gentler introduction to our swap matching methodology
Home page starts with the premise "swapping is possible" then goes straight into "so what's your preferred party" without first laying out why we're asking that particular question (which is rooted in our chosen methodology for matching voters for a swap.

Solutions could be anything from a re-designing the whole on-boarding flow, to just some simple wordsmithing.

baob:
Closes #906, which said

> Home page starts with the premise "swapping is possible" then goes straight into "so what's your preferred party" without first laying out why we're asking that particular question (which is rooted in our chosen methodology for matching voters for a swap.)

Ensures the voter understands what we're doing before we ask the questions.

Refinements welcome. I know the text is long, but we have evidence that the current short text is leading to misunderstandings.

Here's how the home page looks after this change.

image


aspiers:
Some immediate thoughts:

image

aspiers:
> I know the text is long, but we have evidence that the current short text is leading to misunderstandings.

BTW I'd disagree the current text is short!  I think I remember hearing we have this evidence, and I'm certainly convinced that the home page needs to be better, but just curious exactly which evidence you're referring to here?

aspiers:
And also to be clear: fully on board with this effort, thanks a lot for kickstarting it!

baob:
The evidence I meant was people making really odd choices for willing, like they totally misunderstood

aspiers:
Oh yeah, that.  I'd say it's a strong indicator although not bulletproof hard evidence.  Another thing we have seen from looking at Google Analytics yesterday is an **enormous** drop-off from the landing (home) page, which is definitely hard evidence we need to improve this page:

image

aspiers:
That said, that data seems completely wrong for that time range (numbers are all far too low) so I don't know what's going on there.  Worse, GA EOL'd and wiped the entire UA dataset forever in the last ~12 hours :angry:  So we'll never know.

But we see the same story for BE2022:

image

baob:
How about this ?

image

app/views/home/_swap_form.html.haml
```diff
@@ -1,12 +1,14 @@
 .plain-pattern.border-bottom
   .container.text-center
     %p.choose-party
-      My preferred party is
+      What party you would really like to vote for ?
```
aspiers, line 4:
```suggestion
      Which party you would most like to vote for?
```

app/views/home/_swap_form.html.haml
```diff
@@ -1,12 +1,14 @@
...
       = select :user, :preferred_party_id, nil, {}, {form: 'login-form'} do
         %option(disabled=true selected=true) ...
         = options_for_select(@parties.map {|p| [ p.name, p.id ]}, selected: @preferred_party_id)
     %p.choose-party
-      but I'm willing to vote for
+      When we find someone to vote for your party in
+      = election_constituency_other
+      what party could you vote for in exchange ?
```
aspiers, line 11:
```suggestion
      which party could you vote for in exchange?
```

and needs a comma after the constituency.  But this question still feels like it needs simplifying somehow.

app/views/home/_open_pre_elections.html.haml
```diff
@@ -4,12 +4,10 @@
       Make your vote count in the #{election_date_season_type}!
 
     %h2.subdued
-      Can you vote in the #{election_event_choice} on #{election_date_dm}?
-
-    %h2.subdued
-      You can swap votes with someone in #{election_constituency_other} to help
-      both votes count for more. You get to vote for who you really want,
-      and to help someone else do the same!
+      Swap My Vote can help if the party you really want vote for doesn't make sense in
```
aspiers, line 7:
```suggestion
      Swap My Vote can help if voting for your preferred party doesn't make sense in
```

aspiers: (or even if sticking with the original text, there's a missing "to")

app/views/home/_swap_form.html.haml
baob: Not saying it can't be simplified, but the reason it's relatively long is that it includes the explanation of what we're trying to do. And thus delivers the purpose and the question in one go. Rather than a separated explanation, followed by a question later on. (or worse, question followed by explanation, which is what we had).

tdg:
I think this is all on a good track, and would love to review again if we can bring constituency up into the form as discussed on #625

app/views/home/_swap_form.html.haml
aspiers: Yeah agreed, my point is that intuitively I feel there should be some shorter way of combining the purpose and the question.  I could be wrong of course.

tdg:
Here's a take. I'm so close to this it is best that you are reviewing, but I will put it here in case it is of use or opens anything up:

**Double your vote in Thursday's General Election!**

Are you forced into a 'smart vote', a tactical vote where you live?
Find a partner who can vote for your 'heart vote' where they live.

Which party has the best policies and people? 
_This is your 'heart vote' – we will find you a partner who'll cast this vote for you_

Which party makes best sense to vote, where you live? 
_This is your 'smart vote', which you offer to and cast for your partner._

baob:
Now:
image

baob:
> I think this is all on a good track, and would love to review again if we can bring constituency up into the form as discussed on #625

I'd really like to keep that as a separate issue if I can. Reason: this one solves a problem and is (nearly ?) ready to go. If we add the #625 issue, that could cascade into a whole bunch of UX issues that would keep us debating for a couple of days.

Can we get this change in first ?

aspiers:
Just for comparison side by side (well, above/below), here's what we currently have:

image

De-duplicating the first two sentences was definitely a great step in the right direction.  However I'm not so convinced by making the two questions a lot longer; worried this could be a step backwards.  We three are much too close to this to be able to judge, maybe we can quickly get some outside opinions from WhatsApp groups?

baob:
> maybe we can quickly get some outside opinions from WhatsApp groups?

I drafted this two hours ago (before your suggestions came in) and didn't press enter

> So is this good to merge and get onto staging so our testers can check it ? 

So, yes.

aspiers (approved):

PR #944, opened 2024-07-02
Improve mobile phone verification flow

### LINKED ISSUE #166: /user/edit workflow needs refinement

stevebaxter:
The main differences are:

1. Swap the order of the postcode and constituency fields, and make it clear that the postcode is only to find the constituency (this has bugged me for ages)
2. You don't get asked to verify your phone number until the very last moment - when you ask to swap. This is a pain for the user, we want to get them almost to the end of the process (and hooked) before asking for this extra effort.
3. The "Save" and "Verify" buttons are gone from the mobile phone field. It all just happens at the point where you need it to.

Closes #166

### DIFF
--- a/app/controllers/user/swaps_controller.rb
+++ b/app/controllers/user/swaps_controller.rb
@@ -60,8 +60,7 @@ class User::SwapsController < ApplicationController
   def assert_mobile_phone_verified
     return unless @user.mobile_verification_missing?
 
-    flash[:errors] = ["Please verify your mobile phone number before you swap!"]
-    redirect_to edit_user_path
+    redirect_to verify_mobile_path
   end
 
   def assert_has_email
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -45,12 +45,6 @@ class UsersController < ApplicationController
   end
 
   def redirect_path
-    # If the user came from the edit path, and the mobile still needs verification, return there
-    if params[:user] && mobile_set_but_not_verified?
-      flash[:errors] = ["Please click the Verify button to verify the mobile phone number you provided."]
-      return edit_user_path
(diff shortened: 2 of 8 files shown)


stevebaxter:
Video of how it looks here https://www.dropbox.com/s/k9b5t1reoi4x8um/Screen%20Recording%202024-07-02%20at%2022.48.44.mov?dl=0

aspiers (approved):
Thanks this is awesome!  Almost perfect, just one minor point regarding this:

> 2. You don't get asked to verify your phone number until the very last moment - when you ask to swap. This is a pain for the user, we want to get them almost to the end of the process (and hooked) before asking for this extra effort.

Normally I agree, but there is an edge case: when the user has no swaps, they have nothing else to do except for waiting for swaps (see `app/views/user/swaps/_searching_for_swap.html.haml`).  In this case I think it makes sense to get them to verify their number since there is no other call to action available and this saves time later.

aspiers:
I notice that #766 is still an issue in this case.

aspiers:
Merging since that issue is a minor edge case which can be fixed later.

PR #947, opened 2024-07-03
allow verification of mobile whilst waiting for potential swaps

aspiers:
When there are no swaps available, there's no other call to action so it makes sense to capitalise on the waiting time to get the user to verify their mobile, so that they're ready to swap when a partner materialises.

This is still WIP with issues:

- failing tests

- "Verified!" appears the line below the number field, rather than to the right

- probably there's a more elegant way to do it than this `allow_verification` local

### DIFF
--- a/app/views/mobile_phone/_form.html.haml
+++ b/app/views/mobile_phone/_form.html.haml
@@ -2,3 +2,11 @@
   %label.mb-0 My mobile number is
 
   = telephone_field_tag "mobile_phone[full]", mobile_number, size: 14, class: "form-control"
+
+  - if mobile_verified?
+    Verified!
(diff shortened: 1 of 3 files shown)


aspiers:
@stevebaxter I was thinking something like this, what do you reckon?

stevebaxter:
I'm still not sure I would ask for verification until we have something to offer them. I'm easy if you want to add this though.

I wouldn't use exclamation marks in any UI, it can feel a bit excitable (I've been trying to take them all out).

stevebaxter:
I was also trying to get rid of the nasty "Verified" button because we can't integrate it nicely with the phone input (like we have done for postcode).

PR #949, opened 2024-07-03
app.json: make SERVER_HOST optional

aspiers:
Required for cookies to work on review apps, since there we don't set it.


aspiers:
Merging so I can get analytics working tonight.

PR #956, opened 2024-07-03
remove ActiveRecord#update_attributes which was removed in Rails 6.1

aspiers:
https://guides.rubyonrails.org/6_1_release_notes.html#active-record-removals
https://stackoverflow.com/questions/21605186/undefined-method-update-attributes-in-rails

Fixes #918.

PR #957, opened 2024-07-03
fix totally broken error handling on MessageBird verify_create failure

aspiers:
Follow-on from #956 which was masking other issues.


aspiers:
Emergency fix - merging.

PR #963, opened 2024-07-03
fix open-and-voting text

aspiers:


aspiers:
Needs to be ready in a few hours, emergency merging.

Issue #960, opened 2024-07-03
Willing Party profile extended


baob:
Further requirements for the 'Willing Party Profile' feature, originally posted on #905 

## Stability

@aspiers 

> Reopening as I've had to revert this in production due to instability.

This references the problem #951 where the final comment was

> I mistakenly thought this had happened way more times than just twice, so let's massively deprioritize.

## profile_review flag

@aspiers 

> by not showing the page to several thousand existing users who have already set their willing party, I think this greatly diminished the value of the feature since it's only visible to new sign-ups or those who decided to edit their willing party. So the original idea of a new DB flag tracking whether users have properly reviewed their willing party choice still makes a lot of sense to me.

@baob 

> I think it only makes sense if you can get all those users who previously set willing/constituency to visit the review page, perhaps by changing the login flow or something else. Otherwise the new DB field will make next to no difference as they will carry on using their un-reviewed selection.

@aspiers 

> > perhaps by changing the login flow or something else.

> No, simply by making it the default page if they haven't reviewed and haven't already offered or confirmed a swap.

## User flow

@baob 

> It will take a lot of work for me to be confident this won't have a bad interaction with some other feature, and/or introduce new bugs. Particularly with the existing user flow bugs like #824.

@aspiers 

> if you could be super confident of fixing the existing review code so that it applies at least for new users/edits then we could reintroduce it as is, but [lets avoid more] bugs in production

baob:
> if you could be super confident of fixing the existing review code so that it applies at least for new users/edits

@aspiers PR #961 does exactly that.

aspiers:
@baob Thanks for the write-up, this is great!

It's currently too easy for the review to be skipped, either deliberately or accidentally, because the new party is saved even before the review is complete:

Could happen deliberately, e.g. the user clicks their name at top right on the review page => goes back to profile edit, with new party saved.

Could happen accidentally, because there's a bunch of code in between where `review_required` is set and where it's used:

https://github.com/swapmyvote/swapmyvote/blob/f978ed48f1a5b9f579f3f6f3350ea53f807cf058/app/controllers/users_controller.rb#L32-L50

So if there are flash errors, or the phone number fails to update for some reason, then the review redirect won't happen, and then it's too late because as per above the new party is already saved.

[...]

Issue #824, opened 2024-06-18
BUG: willing/preferred party are lost during sign-up


baob:
Steps:

* From home page of SMV, choose willing/preferred party
* you're asked to fill in email and password
* ~make the two passwords different, save~
* ~passwords rejected, so make them match, save~
* (choose constituency)
* now you are asked willing/preferred party again.

aspiers:
This is happening even without doing anything wrong in the flow.

aspiers:
This is breaking the API too.  The API correctly prepopulates the form on the home page prior to login, but after sign-up, those values are lost, just as if the user had manually selected them rather than via prepopulation from the API .

baob:
> This is breaking the API too.

Yes, I fully expected that, didn't test for it.

aspiers:
I think this is how it's _supposed_ to work:

https://github.com/swapmyvote/swapmyvote/blob/3475aed13fde5b82961a620d830df5ab771a5cd4/app/controllers/users/sessions_controller.rb#L12-L31

aspiers:
I have found a change which I'm a bit suspicious of:

https://github.com/swapmyvote/swapmyvote/commit/8a4b078862ce708a93db3356e09257d293899e3f

Admittedly this problem was reported 2 weeks ago and this change is not that old, so maybe there are additional problems lurking behind it.  But there's no getting around the fact that it changes [a `POST` `formaction`](https://github.com/swapmyvote/swapmyvote/commit/8a4b078862ce708a93db3356e09257d293899e3f#diff-e67e56b65817826c0db345725c069e715ac003a4924034910e001015e5acfac2L31) to a [`window.location.assign`](https://github.com/swapmyvote/swapmyvote/commit/8a4b078862ce708a93db3356e09257d293899e3f#diff-7ad9acaf8f76ede9d0ba0fcf552faf830ff5b925b9b080b2be0bcede863e5f56R43) which I assume results in a `GET` not a `POST`.  And after this `window.location.assign` I can't see any evidence of the user's party choices persisting on the sign-up page either in `params` or `session`.

Bearing mind the above comment, it seems clear that this whole process relies on a `POST` not a `GET` to preserve the party choices, so I think that's a good place to start investigation.  Perhaps reverting to the commit before this one and seeing whether the choices get passed to the sign-up page (and beyond) is the best next step, but I've run out of time.

@stevebaxter This is your wheelhouse for sure, but if @baob gets to it first then great.

[...]

PR #965, opened 2024-07-03
user/edit: show errors if willing/preferred parties missing

aspiers:
Currently if you sign up, you land on `/user/edit` with empty parties due to bug #824.  If you then hit Save, it goes nowhere but doesn't explain why, so add flash errors for this.

However, this increases the risk of the willing party review being accidentally skipped (i.e. if preferred party was left blank), for reasons described here:

https://github.com/swapmyvote/swapmyvote/issues/960#issuecomment-2207404365

Given the last 24 hours, I'm nervous about this change and it needs careful testing.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -45,6 +45,13 @@ class UsersController < ApplicationController
       end
     end
 
+    unless @user.willing_party
+      (flash[:errors] ||= []).append "You must state which party you are willing to vote for."
+    end
+    unless @user.preferred_party
+      (flash[:errors] ||= []).append "You must state which party you would prefer to vote for."
+    end
+
     no_flash_errors = (!flash[:errors] || flash[:errors].size.zero?)
 
     if @user.valid? && no_flash_errors && review_required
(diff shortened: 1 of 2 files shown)


baob (approved):
I tried to break it, even using scenarios that should cause problems listed in #960, but it all looks good to me. If you blank email and change willing at the same time, the update is rejected so on the correcting submit the voter has to change willing again, and so the change is captured and the redirect to /user/review happens correctly.

app/controllers/users_controller.rb
```diff
@@ -45,6 +45,13 @@ def update
       end
     end
 
+    unless @user.willing_party
+      (flash[:errors] ||= []).append "You must state which party you are willing to vote for."
```
stevebaxter, line 49: Consider not ending a message with a "." unless it contains multiple sentences.

stevebaxter (approved):
I've done some testing, it all seems to work for me!

app/controllers/users_controller.rb
aspiers: It's pretty tricky to implement given the possibility of what can go into `flash[:errors]`.  Not worth the effort here.

stevebaxter: Each one is in a separate alert, so they stand alone.

aspiers: Ah you mean a separate red box?

stevebaxter: Yes

PR #969, opened 2024-07-04
rake task to send email reminders to get swapping

### LINKED ISSUE #925: Email reminder to choose a swap
For any user who hasn't chosen a swap OR been offered a swap, we should be able to send out a email reminder (just once per user) prodding them into action, triggered via a Rake task which can be run hourly from the Heroku scheduler.  We can use the `SentEmails` model to track when this was sent, to ensure that we don't spam users.  It should be sent if the current time is after `User.created_at + 0.25 * ENV['SWAP_EXPIRY_HOURS']`.

aspiers:
For now, emails are only sent once per user.

Fixes #925.


aspiers:
Emergency use case, needs to be sent YESTERDAY, so merging.

PR #971, opened 2024-07-04
send reminders to review pending swaps

### LINKED ISSUE #923: Email reminder to review and accept or decline offered swap
This should tell the user how many hours left they have before the swap expires.  We should have a rake task which runs hourly via the Heroku scheduler, and sends this email if the offered swap is more than 50% of the way to expiry (but see also #924, since currently a swap could suddenly jump from 48% to 100%).

aspiers:
Will fix #923, but not ready yet.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -315,17 +315,19 @@ class User < ApplicationRecord
     sent_emails.create!(template: SentEmail::REMINDER_GET_SWAPPING)
   end
 
-  def send_pending_swap_reminder_email
-    # FIXME: check SentEmail to avoid duplicates
-    # return if ...
-    self.pending_swap_reminder_email = true
+  def send_pending_swap_reminder_email(other_user)
+    template = SentEmail::REMINDER_PENDING_OFFER
+    already_sent = sent_emails.where(template: template).first
+    return if already_sent
+
     save
-    UserMailer.reminder_to_accept_swap(self).deliver_now
-    sent_emails.create!(template: SentEmail::REMINDER_PENDING_OFFER)
+    UserMailer.reminder_to_accept_swap(self, other_user).deliver_now
+    sent_emails.create!(template: template)
   end
 
   def send_vote_reminder_email
     # FIXME: check SentEmail to avoid duplicates
+    # Errrm, looks like we have two ways to check if this mail was sent ?
     return if sent_vote_reminder_email
     self.sent_vote_reminder_email = true
     save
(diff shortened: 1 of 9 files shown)


aspiers:
@baob As discussed with @tdg would be useful to have this ready to run around 12pm if you don't have more urgent things to work on.  Ask him for help prioritising as my brain has stopped now.  This PR is well on the way but needs updating to copy a few of the newer changes from #969 and #970, as I drafted this while I was ~half way through #969.

app/models/user.rb
```diff
@@ -315,17 +315,19 @@ def send_get_swapping_reminder_email
     sent_emails.create!(template: SentEmail::REMINDER_GET_SWAPPING)
   end
 
-  def send_pending_swap_reminder_email
-    # FIXME: check SentEmail to avoid duplicates
-    # return if ...
-    self.pending_swap_reminder_email = true
+  def send_pending_swap_reminder_email(other_user)
+    template = SentEmail::REMINDER_PENDING_OFFER
+    already_sent = sent_emails.where(template: template).first
```
aspiers, line 320: Completely unimportant, just for my curiosity wouldn't `sent_emails.find_by_template(template)` do the trick here?

app/models/user.rb
```diff
@@ -315,17 +315,19 @@ def send_get_swapping_reminder_email
...
     save
-    UserMailer.reminder_to_accept_swap(self).deliver_now
-    sent_emails.create!(template: SentEmail::REMINDER_PENDING_OFFER)
+    UserMailer.reminder_to_accept_swap(self, other_user).deliver_now
+    sent_emails.create!(template: template)
   end
 
   def send_vote_reminder_email
     # FIXME: check SentEmail to avoid duplicates
+    # Errrm, looks like we have two ways to check if this mail was sent ?
```
aspiers, line 330: Yes, there's the old way which we should replace with this "new" `SentEmail` model, otherwise we have an ugly mess of email stuff in the `User` model.  My second commit was a totally unfinished attempt to clean that up, although we can defer that to a separate PR because not changing it for this GE.

app/models/user.rb
baob: yes, but it doesn't read as nice :)

aspiers: I like the temporary variable, main point was about `find_by_template` instead of `.where(...).first`

baob: I guess I was influenced by seeing the other `SentMail.where` . Rubocop has a preference between `.find_by_X("Y")`  and `find_by(X: "Y")` and and I can't remember which way round it is.

aspiers: Both work!

aspiers: Oh sorry, you meant Rubocop.

app/models/user.rb
aspiers: Filed as #982.

PR #974, opened 2024-07-04
Restore button for mobile verification

stevebaxter:


aspiers (approved):

app/views/mobile_phone/_form.html.haml
```diff
@@ -1,4 +1,12 @@
...
 
   = telephone_field_tag "mobile_phone[full]", mobile_number, size: 14, class: "form-control"
+  - if mobile_number.present?
+    - if mobile_verified?
+      (verified)
+    - else
+      = button_tag "Verify", formaction: verify_mobile_path,
+        formmethod: :get, class: "btn btn-secondary"
+  - else
+    = submit_tag("Save", class: "btn btn-secondary")
```
aspiers, line 12: I'm still worried this might cause Save button to appear in places we don't want it.

aspiers (approved):

aspiers (approved):

app/views/mobile_phone/_form.html.haml
stevebaxter: This is exactly the same code as before though. All I've done is reverted the change.

aspiers: Discussed and @stevebaxter tested it works fine - it does appear but it does the right thing.

PR #978, opened 2024-07-04
Fix issue with willing/preferred party being lost on sign up

### LINKED ISSUE #824: BUG: willing/preferred party are lost during sign-up

stevebaxter:
We need to submit the willing and preferred party as a form or they won't be carried through to the next page.

Closes #824.

### DIFF
--- a/app/views/home/_swap_form.html.haml
+++ b/app/views/home/_swap_form.html.haml
@@ -1,22 +1,24 @@
 .plain-pattern.border-bottom
   .container.text-center
-    %p.choose-party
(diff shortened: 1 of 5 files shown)


stevebaxter:
Not sure why hound is upset, that is the original code

aspiers (approved):

Issue #922, opened 2024-07-01
Introduce "wizard" questionnaire for onboarding


aspiers:
I just spoke with @tdg and we came up with some really interesting ideas around the "wizard" idea which might not be too much work to implement, and which we both think could be a huge improvement.  Briefly:

- Only show one question at a time.
- First page should briefly "sell" the idea of vote swapping, e.g. via the "double your vote" lens.  It should then pose a single question asking for the user's constituency (or perhaps even have a "Get started" button which leads to a separate page for this question).
- Second is preferred, focusing on idealism, e.g. "if you could vote with your heart, which party would it be for?" or similar
- Third is willing, focusing on pragmatism. But by this point we already know constituency and preferred, so we can:
  - Show the bar chart for the user's constituency
  - For every party, show the potential impact of voting for that party locally (exactly like you are doing in #905, but for all parties) 
  - Ask the user to choose a willing party, phrasing the question in a way which is highly contextualised to the user's personal situation.  This should make it a lot easier to understand, e.g.
    > Which party would you be willing to vote for in $MY_CONSTITUENCY, in return for someone elsewhere voting for $MY_PREFERRED where it could make a bigger difference?" 
  - Get the user to choose one of these parties with a single click (better than a dropdown)

If the user later edits constituency or preferred from the profile, it could (should?) take them back to this choice-of-willing page.  If they want to edit willing, they should only be able to do it from this choice-of-willing page which shows the impact of their change of choice.

aspiers:
Admittedly this is a big stretch with limited time left, may not be possible.

aspiers:
Also, if any fields are prepopulated via API, then those pages would simply be skipped (but ideally with an option to go back and change previous settings, regardless of whether those came from the user or API).

aspiers:
At this stage, it's unrealistic to contemplate doing this with Javascript.  We'd just need separate Rails routes/views for each step.

baob:
It's good. It's the kind of thing Tom and I discussed recently and was also discussed post BE2022 as part of the 'UX overhaul' thing. But

* given my availability
* given the amount of time we spend negotiating words, never mind workflow within the wizard

I don't see it happening this election.

baob:
We need to get both of these done as a pre-requisite

* Bug #824 
* Merge PR #897

PR #1011, opened 2024-10-25
constituency first

baob:
When complete, this will close https://github.com/swapmyvote/swapmyvote/issues/625 and possibly #922 too. 

Most of the work done to move the constituency-first commits over, making constituency the first question asked during registration.

Still work to do on the postcode lookup 

# Minor style issues

(I have been unable to resolve these, may need @stevebaxter 's help)

* Postcode select on constituency page (home page) is misaligned

# Minor Issues

* specs (some commented out) still want to see parties set by the constituency (first) page rather than constituencies
* there are no specs to require constituencies on the new, second, constituencies page. 
* constituencies are not in alpha order at time of writing, may change with postcode lookup fixes above.

### DIFF
--- a/.haml-lint.yml
+++ b/.haml-lint.yml
@@ -29,9 +29,8 @@ linters:
   EmptyScript:
     enabled: true
 
-  # Seems to default to true anyway, but uncommenting it interferes with the TODO file
-  # HtmlAttributes:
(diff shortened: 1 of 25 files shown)


baob:
Closed see #1024 instead

Issue #1023, opened 2024-11-02
Resolve hound complaint about Rails/I18nLocaleTexts found in .rubocop..yml


baob:
Following the suggestion in the text

> Error: unrecognized cop or department Rails/I18nLocaleTexts found in .rubocop.yml
> Did you mean `Rails/I18nLocaleAssignment`?

Does not not lead to any easy resolution, just the critical errors 

```
Offenses:

app/mailers/user_mailer.rb:15:36: C: Move locale texts to the locale files in the config/locales directory.
    mail(to: @user.email, subject: "Welcome to Swap My Vote")
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^
```

Discovered while working on PR #1011 

It looks like, though this [passed on first push](https://github.com/swapmyvote/swapmyvote/actions/runs/11766497105), the version of rubocop selected is no longer supported. Pick a version from here

http://help.houndci.com/en/articles/2461415-supported-linters

---

## Stuck between a rock and a hard place. 

Currently, you get this error from hound CI

```
Some files could not be reviewed due to errors:

Error: unrecognized cop or department Rails/I18nLocaleTexts found in .rubocop.yml
Did you mean `Rails/I18nLocaleAssignment`?
```

However if you remove the offending line in .robocop.yml it appears that it is actually necessary to have that cop configured and turned off.

On CI rubocop (not hound), and local rubocop on you get

```
Offenses:

app/mailers/user_mailer.rb:15:36: C: Rails/I18nLocaleTexts: Move locale texts to the locale files in the config/locales directory.
    mail(to: @user.email, subject: "Welcome to Swap My Vote")
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^
app/mailers/user_mailer.rb:52:36: C: Rails/I18nLocaleTexts: Move locale texts to the locale files in the config/locales directory.
    mail(to: @user.email, subject: "Your vote swap preference is in demand!")
```    
    
Looks like hound CI and rubocop fundamentally disagree with eachother.

---

## Probable solutions

* Turn off the rubocop feature of hound CI - don't need it we have a rubocop check in our own CI workflow.
* find a really old version of rubocop where rubocop and hound no longer disagree.

@aspiers Would appreciate your comments.

aspiers:
#1027 was merged which fixes this.

PR #1024, opened 2024-11-02
syv followup constituency first

### LINKED ISSUE #625: Ask user for constituency before showing them swaps and before asking their 'willing' party

baob:
This will close #625 and possibly #922 too.

Most of the work done to move the constituency-first commits over, making constituency the first question asked during registration.

The API still works.

User edit (constituency) confirmed working

### DIFF
--- a/app/views/home/_constituency_form.html.haml
+++ b/app/views/home/_constituency_form.html.haml
@@ -7,5 +7,7 @@
                           selected: default_constituency_ons_id,
                           prompt: true
 
+  = render partial: "users/postcode_field"
+
   %p.text-center
     = submit_tag("Save", class: "btn btn-primary")
--- a/app/views/home/_constituency_form.html.haml
+++ b/app/views/home/_constituency_form.html.haml
@@ -11,4 +11,4 @@
   = render partial: "users/postcode_field"
 
   %p.text-center
-    = submit_tag("Save", class: "btn btn-primary")
+    = submit_tag("Next: Choose Parties", class: "btn btn-primary")
--- a/app/views/home/_constituency_form.html.haml
+++ b/app/views/home/_constituency_form.html.haml
@@ -1,0 +1,11 @@
+= form_with url: pre_login_path, id: "login-form", local: true do |f|
+
+  %p.ui-widget
+    My constituency is
+    = f.collection_select :constituency_ons_id, @constituencies,
+                          :ons_id, :name,
+                          selected: @default_constituency_ons_id,
+                          prompt: true
+
+  %p.text-center
+    = submit_tag("Save", class: "btn btn-primary")
--- a/app/views/home/_constituency_form.html.haml
+++ b/app/views/home/_constituency_form.html.haml
@@ -1,6 +1,7 @@
 = form_with url: pre_login_path, id: "login-form", local: true do |f|
 
-  %p.ui-widget
+
+  %p.choose-constituency
     My constituency is
     = f.collection_select :constituency_ons_id, constituencies,
                           :ons_id, :name,
--- a/app/views/home/_constituency_form.html.haml
+++ b/app/views/home/_constituency_form.html.haml
@@ -1,4 +1,4 @@
-= form_with url: pre_login_path, id: "login-form", local: true do |f|
+= form_for @user, url: pre_login_path, id: "login-form", local: true do |f|
 
   .d-flex.flex-column.align-items-center
 
--- a/app/views/home/_constituency_form.html.haml
+++ b/app/views/home/_constituency_form.html.haml
@@ -1,4 +1,4 @@
-= form_for @user, url: pre_login_path, id: "login-form", local: true do |f|
+= form_for user, url: pre_login_path, id: "login-form", local: true do |f|
 
   .d-flex.flex-column.align-items-center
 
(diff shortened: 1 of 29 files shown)


baob:
> Some files could not be reviewed due to errors:

Raised issue #1023 to address this

app/views/home/_constituency_form.html.haml
```diff
@@ -0,0 +1,14 @@
+= form_with url: pre_login_path, id: "login-form", local: true do |f|
```
stevebaxter, line 1: So I think here you want to do:

```
.d-flex.flex-column.justify-content-center
```

See here for example:

https://github.com/swapmyvote/swapmyvote/blob/8ca61f9a1ba59e87ee61d4e50ad55a1f28339ecf/app/views/user/swaps/_swap_profile.html.haml#L1

This should then lay out all the divs vertically and centred horizontally. You can read about the built-in BS flex classes here:

https://getbootstrap.com/docs/5.3/utilities/flex/

app/views/home/_constituency_form.html.haml
```diff
@@ -0,0 +1,14 @@
...
+
+
+  %p.choose-constituency
```
stevebaxter, line 4: I'm not sure you want to use paragraphs for this, instead I would use divs with standard padding so each div gets its own padding. At the moment it isn't quite right, as the postcode field doesn't get the right padding.

app/views/home/_constituency_form.html.haml
```diff
@@ -0,0 +1,14 @@
...
+    My constituency is
+    = f.collection_select :constituency_ons_id, constituencies,
+                          :ons_id, :name,
+                          selected: default_constituency_ons_id,
+                          prompt: true
+
+  = render partial: "users/postcode_field"
```
stevebaxter, line 11: You might need to put this in a div to make sure it is rendered with the right padding.

app/views/home/_constituency_form.html.haml
stevebaxter: Sorry, try this:
```
= form_with url: pre_login_path, id: "login-form", local: true do |f|

  .d-flex.flex-column.align-items-center

    %p.choose-constituency
      My constituency is
      = f.collection_select :constituency_ons_id, constituencies,
                            :ons_id, :name,
                            selected: default_constituency_ons_id,
                            prompt: true

    = render partial: "users/postcode_field"

    %p.text-center
      = submit_tag("Next: Choose Parties", class: "btn btn-primary")

```
The spacing still isn't right because the CSS is a bit inconsistent, this will centre it though.

baob:
# Preview 

image


image

app/views/users/_postcode_field.html.haml
```diff
@@ -1,11 +1,12 @@
 %label.mb-0 Or find my constituency using my postcode
 
-.form-group.form-group-short
-  .input-group
-    %input.form-control#txt-postcode{ type: "text", size: 10, minlength: 5,
-                                      maxlength: 9, spellcheck: "false" }
-    .input-group-append
-      %input.btn.btn-secondary#btn-postcode{ type: "button", value: "Search" }
+.d-flex.flex-row.justify-content-center
```
stevebaxter, line 3: I don't think this is right. We don't want to always centre the postcode field. Instead I think you need to do what I suggested here?

https://github.com/swapmyvote/swapmyvote/pull/1024/files#r1835562684

baob:
> Some files could not be reviewed due to errors:
> 
> Error: unrecognized cop or department Rails/I18nLocaleTexts found in .rubocop...
> Error: unrecognized cop or department Rails/I18nLocaleTexts found in .rubocop.yml
> Did you mean `Rails/I18nLocaleAssignment`?

This is really not straightforward to fix, certainly not as part of this PR. See https://github.com/swapmyvote/swapmyvote/issues/1023

tdg:
These look great. 

The change of call to action is radical but probably overdue. 

Can I suggest the second screen have as its button whatever the next step is, in the new onboarding journey, for example: 
"Next: find my swap"
or
"Next: verify"

baob:
@tdg The sequence is now

image

image

image

tdg:
Nice, sign up is great

tdg:
Maybe the last one should say "confirm and see swaps"

aspiers (approved):

PR #1046, opened 2026-08-23
M3 part 1: entry-form API and postcode/constituency lookup

stevebaxter:
First of two PRs for M3 (home / landing). This one lands the reference data the home page renders from, plus the two controls that replace the jQuery postcode helper. **No screens yet, and nothing is routed to React** — the two-step entry wizard and the five phase screens follow in part 2.

## API

All unauthenticated and ungated, on the `Api::V1::BaseController` M2 established — the entry form comes before sign up.

| endpoint | returns |
| --- | --- |
| `GET /api/v1/parties` | every party, alphabetically |
| `GET /api/v1/constituencies` | every constituency, keyed on the ONS GSS code |
| `GET /api/v1/election` | the election, and the prose built from it |

### Why an election endpoint

The plan didn't anticipate this one, so it's worth flagging. The home page's headings are assembled from about a dozen `ApplicationHelper` strings — `election_date_season_type`, `general_election?`, `election_constituency_other`, `hide_polls?`, `by_election_constituencies_as_sentence` — each derived from `ELECTION_DATE` / `ELECTION_TYPE` env vars and from how many `OnsConstituency` rows exist.

Reimplementing that date/season logic in TypeScript would fork the source of truth and drift the moment either input changed, so Ruby computes the strings and the API sends them whole.

It's a separate endpoint rather than part of the session payload because the two have opposite lifetimes: the session is re-fetched on a 60s poll, while the election is fixed for the life of a deploy. `lib/referenceData.ts` fetches it once and caches it indefinitely.

## Frontend

- `lib/postcodes.ts` — postcodes.io lookup, replacing `app/javascript/packs/postcodesHelper.js`. Still calling postcodes.io directly rather than tacticalvote's `/api/lookup`, per "Code sharing with tacticalvote" in the plan.
- `PostcodeLookup` — the field, its Search button and its error.
- `ConstituencyAutocomplete` — name in, ONS code out, via a native `<datalist>` rather than a library.

Behaviours carried over from the jQuery helper that would be easy to lose, each with a test:

- **Enter searches, it does not submit.** The postcode field sits inside the entry form and the user usually hasn't filled in the rest of it.
- **A postcode outside our constituencies is an error *and* clears the selection**, rather than leaving a stale one behind.
- **400/404 show postcodes.io's own wording** ("Postcode not found"); any other status is reported as a service error with the body, so an outage stays distinguishable from a typo.
- The autocomplete syncs when the postcode lookup fills it in — the legacy helper set that input directly.

The postcode is used and discarded: never sent to our backend, never stored. The copy under the field still says so.

## Note on the legacy pack

`postcodesHelper.js` and its HAML callers are untouched and still serving `/`. Deletion waits for cutover, per the plan.

## Checks

- `yarn lint`, `yarn typecheck` clean; `yarn test` 124 passing (20 files), 26 of them new here.
- `bundle exec rspec` 427 examples, 16 failures — the same 16 that fail on `master` (legacy HAML view specs dying on a missing Webpacker manifest, plus the mailer/snapshot ones). All 26 new request specs pass.
- `rubocop` clean on the new Ruby.

### DIFF
--- a/app/controllers/api/v1/constituencies_controller.rb
+++ b/app/controllers/api/v1/constituencies_controller.rb
@@ -1,0 +1,14 @@
+module Api
+  module V1
+    # The constituencies the site is currently running swaps in — two for a
+    # by-election, the full list for a general election. Reference data, same
+    # as parties: no auth, no phase gate.
+    class ConstituenciesController < BaseController
+      def index
+        render json: ConstituencySerializer.new(
+          OnsConstituency.all.order(:name)
+        ).to_h
+      end
+    end
+  end
+end
--- a/app/controllers/api/v1/election_controller.rb
+++ b/app/controllers/api/v1/election_controller.rb
@@ -1,0 +1,13 @@
+module Api
+  module V1
+    # The election being run for — dates, type, and the prose the SPA builds
+    # its headings from. Immutable for the life of a deploy, so it is a
+    # separate endpoint the client can cache indefinitely rather than part of
+    # the session payload, which is re-fetched on a poll.
+    class ElectionController < BaseController
+      def show
+        render json: ElectionSerializer.new(ElectionPresenter.new(self)).to_h
+      end
+    end
+  end
+end
--- a/app/controllers/api/v1/parties_controller.rb
+++ b/app/controllers/api/v1/parties_controller.rb
@@ -1,0 +1,11 @@
+module Api
+  module V1
+    # The parties a user can pick as their preferred or willing party.
+    # Reference data: no auth, no phase gate, changes only when we seed it.
+    class PartiesController < BaseController
+      def index
+        render json: PartySerializer.new(Party.all.order(:name)).to_h
+      end
+    end
+  end
+end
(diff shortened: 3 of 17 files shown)


stevebaxter:
### Follow-up decided: `ConstituencyAutocomplete` moves to react-bootstrap-typeahead in part 2

The version in this PR uses a native `<datalist>`. Its dropdown renders as browser chrome — outside the page — so it cannot be positioned, styled or measured by us, and it renders in the wrong place in some browser contexts. It also can't provide the legacy widget's show-all toggle.

Part 2 replaces it with [`react-bootstrap-typeahead`](https://github.com/ericgio/react-bootstrap-typeahead), which renders the list as real DOM. That closes the two remaining gaps against the jQuery combobox in `app/assets/javascripts/autocomplete.js`:

- the **show-all toggle button** (the legacy widget runs at `minLength: 0`)
- **clearing the postcode field when a constituency is picked by name** — legacy `autocompleteselect` does `$("#txt-postcode").val("")`, since the two inputs are two ways of answering the same question

The behaviours already ported here — substring filtering, exact-match-to-select, and clearing a half-typed name on blur (`_removeIfInvalid`) — carry over, and their tests should survive the swap largely unchanged since they assert through the label and value rather than the markup.

Merging this as-is is fine: nothing is routed to React yet, so the datalist never reaches a user.

PR #1066, opened 2026-09-07
Clear the Dependabot security alerts that Rails 6.1 does not block

stevebaxter:
113 of the 129 open Dependabot alerts are fixable without changing the app's Rails version. This does all of them.

## npm — all 36 alerts cleared

`yarn npm audit` is clean afterwards.

- react-router 7.18.1 → 7.18.3, browserslist → 4.28.9, postcss → 8.5.28, undici → 7.29.1.
- **node-gyp 10.1.0 → 13.0.2**, which is what dragged in the vulnerable tar/cacache/socks chain. `brace-expansion`, `cross-spawn`, `glob`, `ip-address` and `minimatch` leave the tree entirely with it.
- **postcss-preset-env 6 → 11, postcss-import 12 → 17, postcss-flexbugs-fixes 4 → 5.** The 2019-era versions pinned postcss 7, which has no patched release, so bumping them was the only way to clear those alerts. `vite build` still produces the same asset set.

## rubygems — 77 of 93 alerts cleared

nokogiri, rack, net-imap, puma, faraday, oauth, oauth2, jwt, concurrent-ruby, websocket-driver, loofah, rails-html-sanitizer, crass, mail, msgpack, addressable, bcrypt, jquery-ui-rails, rexml and uri.

`crass` is not one Dependabot has filed — `bundler-audit` reports it (GHSA-wwpr-jff3-395c), and it is picked up here for free.

### Two bumps pull a dependent along with them

- **jwt 2.2.3 → 3.2.0 forces messagebird-rest 3.1.1 → 5.0.0.** 3.1.1 pins `jwt ~> 2.2.3`, and no messagebird-rest release before 5.0.0 allows a patched jwt, so there is no smaller move. The `verify_create` / `verify_token` / `verify_delete` signatures `SwapMyVote::MessageBird` calls are unchanged in 5.0.0, and the mobile-verification specs pass.
- **puma 6.4.3 → 8.0.2.** `config/puma.rb` uses nothing Puma has dropped, and 8.0.2 still ships `lib/rack/handler/puma.rb` for Rails 6.1. ⚠️ It could not be booted locally: puma lives in the `production` group, which will not install without libpq for `pg`. CI and staging are its first real exercise.

## Two fallouts from the bumps, fixed here

- **concurrent-ruby 1.3.5 dropped its implicit `require "logger"`,** which Rails 6.1's `ActiveSupport::LoggerThreadSafeLevel` relies on. The require goes in `config/boot.rb` rather than `config/application.rb` because `bin/rails` and `bin/rake` load railties straight after boot, before `application.rb` is read — `bin/rake spec`, which is what CI runs, dies without it. (`rspec` alone would have passed with it in `application.rb`; this only surfaced when booting the server for E2E.)
- **rack 2.2.24 reads a new `RACK_MULTIPART_BUFFERED_UPLOAD_BYTESIZE_LIMIT` env var while handling a request,** so specs that stub `ENV#[]` for a single key without `and_call_original` first now fail whenever the seed puts them after that read. Three were missing it; the API request specs already had the pattern.

## Left open — 16 alerts, all gated on `gem "rails", "~> 6.1"`

The actionpack, actionview, activerecord, activestorage and activesupport advisories (2 of them critical), plus devise (5.0.4 needs `railties >= 7.0`) and sqlite3 (2.x needs `activerecord >= 7.0`). Clearing those means upgrading Rails off 6.1, which belongs in its own change.

## Verification

- `bundle exec rspec` — 579 examples, 10 failures, across seeds 61250, 42164 and 1234. The same ten fail on `master` in this environment, from a missing `SERVER_HOST` for the mailers and a stale Vite manifest snapshot; the failure set is byte-identical before and after.
- `bin/rake lint` (rubocop + scss-lint + haml-lint), `yarn lint`, `yarn typecheck`, `yarn test` (429 tests) — all green.
- `yarn e2e` — 25 passed, against a locally booted stack.
- `vite build` — succeeds.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


stevebaxter:
@baob please can you have a look at this if you get a chance? Looks like we should also upgrade to Rails 7, any gotchas on that you know about or shall I just let Claude get on with it?

PR #1073, opened 2026-09-08
M8: vote and share in React

stevebaxter:
Ports the last two user-facing screens before cutover — `User::VoteController` ("I've voted") and `User::ShareController` — and rewrites the SPA's social sharing.

As with every milestone, **the legacy HAML site is untouched and still live**. `/user/vote` and `/user/share` keep serving HAML; React previews under `/app/vote` and `/app/share` until the single cutover at M9.

Design: [`docs/superpowers/specs/2026-09-08-m8-vote-and-share-design.md`](docs/superpowers/specs/2026-09-08-m8-vote-and-share-design.md)

## API

`POST /api/v1/vote` — write-only (`resource :vote, only: [:create]`). There is deliberately no `GET`: `hasVoted` (new on `UserSerializer`) and the willing party ride on the session payload, and the partner's name comes from `GET /api/v1/swap`, which discloses it only on a confirmed swap — exactly this endpoint's gate. It answers with the session payload, so the chrome updates in one round trip.

## Sharing was rewritten, not ported

The React site now offers **WhatsApp, Bluesky and the OS share sheet** (`navigator.share`, with a `react-share` fallback row where it is unavailable), taking tacticalvote's `ShareBox` shape rather than the legacy `_social` partial's Facebook/X/email popups.

This also lands the pattern tacticalvote prototyped but never merged (`proto-react-share-fallback`), raised back to them as [forwarddemocracy/tacticalvote#971](https://github.com/forwarddemocracy/tacticalvote/issues/971). The hydration flicker that stalled it there is a Next SSR artifact and cannot occur in a fully client-rendered SPA, so the capability check is a plain render-time call.

## Three deliberate divergences

1. **The vote screen requires a *confirmed* swap.** Legacy `require_swap` tests `swapped?` — any swap — and the view then prints the partner's real name, contradicting the disclosure rule M7 put in `SwapPartnerDetailSerializer`. The loose guard is unreachable anyway: `shared/_go_vote` only links there when `swap_confirmed?`. The new `require_confirmed_swap!` answers 409.
2. **`POST` is idempotent.** Legacy `#create` mails `partner_has_voted` unconditionally on every call. A JSON endpoint is a double-click away from a duplicate email; the legacy screen never intended one, it just had nothing guarding it.
3. **X is dropped from the React site.** A product decision, not a port. The HAML site keeps its own X button until M9 removes it wholesale.

`GoVote` now links to `/app/vote` as a router `<Link>` — the last thing in the SPA still pointing at a HAML screen it had replaced.

## Tests

`spec/requests/api/v1/vote_spec.rb`, component tests under `app/frontend/components/share/` and `app/frontend/pages/`, `playwright-tests/vote.spec.ts`, and two new entries in `accessibility.spec.ts`.

- Biome lint clean (191 files) · `tsc --noEmit` clean
- Vitest **449 passed**
- RSpec **587 examples, 0 failures** (4 pre-existing pending)
- Rubocop clean (285 files)
- Playwright: M8's four new tests all pass; axe reports no WCAG A/AA violations on either new screen

## Two notes for reviewers

**The local E2E suite is flaky, independently of this branch.** Failures land on pre-existing spec files timing out on `page.goto`, and the count moves between runs — running *only* the pre-existing specs, with M8's excluded, produced *more* failures (8) than the full suite (5). One dev-mode Rails server saturated by 4 parallel workers, with `seedProfileUser` spawning a Rails boot per spec file at module load. CI's `retries: 2` likely masks it. Worth its own issue; not addressed here, since it means changing shared seeding infrastructure.

**`.env.test` is not in the repo** — only `.env.test.example`, and it is not gitignored. Without it `SERVER_HOST` is unset and every mailer-touching spec fails with "Missing host to link to!". Each checkout currently needs it created by hand. Also worth fixing separately.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


stevebaxter:
**Correction.** This PR's description said `GoVote` was "the last thing in the SPA pointing at a HAML screen it had replaced". That was wrong, and I've corrected it in the plan on this branch (`7de3d83d`).

Manual testing turned up two nav links that had been pointing at already-ported legacy paths for several milestones:

- the user menu's **"Edit profile"** → `/user/edit` (ported to `/app/profile` in M4 and `/app/mobile` in M6)
- the **brand logo** → `/` (ported to `/app/home` in M3)

Since the nav is on every SPA screen, both dropped the user into the Bootstrap 4 HAML site with no route back. `Navigation.test.tsx` asserted the old hrefs, so the suite was pinning the bug rather than catching it.

The claim originated in the M6 notes and I repeated it here without checking. Fixed separately in #1076, branched from `master` — it predates this work and shouldn't hold up this review.

